PR #10, opened 2021-02-01
Add support for dropping rcVersion from annotation

aiyengar2:
This commit adds support for dropping rcVersions from any generic set of annotations that were defined under helm/annotations.go.

Currently, this only adds support for such a use case for the annotation catalog.cattle.io/auto-install.

### DIFF
--- a/pkg/helm/annotations.go
+++ b/pkg/helm/annotations.go
@@ -1,0 +1,27 @@
+package helm
+
+import (
+	"strings"
+)
+
+// ExportAnnotation is an annotation that is modified on an export
+type ExportAnnotation interface {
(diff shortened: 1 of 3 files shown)


aiyengar2:
Fixes https://github.com/rancher/charts/pull/958#discussion_r567947598.

The expectation for developers now is to append the correct `releasedCandidateVersion` to the crd-install annotation if they specify a specific version to match against. In production, it will automatically drop the rcVersion from the annotation, if it exists.

This only affects the rancher-istio chart since it is the only chart that actually specifies a rcVersion, whereas the others just use `=match`.

brendarearden (approved):
LGTM

cbron (approved):

PR #15, opened 2021-02-05
Add better support for release workflows, update docs, and some misc changes

aiyengar2:
This PR clumps together some of the changes that need to be addressed to the charts-build-scripts. I'll follow up with `make docs` commands on each of the charts branches.

Changes:
- Pretty print the releaseCandidateVersions selected during a `make sync` when you drop `releaseCandidateVersions` (Live Branch)
- Add `make release` to the Staging Branch to help guide the process of automatically cleaning it up during a release.
- Port https://github.com/rancher/partner-charts/pull/74 changes here (just adds a `_config.yml` file)
- Update docs for release workflow + versioning stuff
- Bump the version of the scripts so we can tag it when we port it over to the charts branches.
- Add .DS_Store to .gitignore


StrongMonkey (approved):
I think you are pretty confident on the codebase so I will just give LGTM :smile:

pkg/sync/compare.go
```diff
@@ -81,18 +82,19 @@ func CompareGeneratedAssets(rootFs billy.Filesystem, newCharts, newAssets, origi
 			for _, f := range fileInfos {
 				chartVersion := f.Name()
 				splitChartVersion := strings.Split(chartVersion, "-rc")
-				chartVersionWithoutRC := strings.Join(splitChartVersion[:len(splitChartVersion)-1], "")
-				latestRCSeenSoFar, ok := latestRC[chartVersionWithoutRC]
+				chartVersionWithName := fmt.Sprintf("%s/%s", chart, chartVersion)
+				chartVersionWithNameWithoutRC := fmt.Sprintf("%s/%s", chart, splitChartVersion[0])
```
cbron, line 86: Are you confident they will all have RC's ? And that this [0] doesn't hit out-of-bounds.

aiyengar2: Even if they don't have an RC, splitChartVersion[0] will just contain the packageVersion of the chart. However, this isn't an edge case I've fully accounted for; I'm not sure how not having RCs will impact https://github.com/rancher/charts-build-scripts/blob/5d6fca48e3d5366200999599c9d60f24a3318ddd/pkg/sync/compare.go#L92-L94.

What should the expected behavior be when RC exist but a packageVersion without the RC also exists on the staging branch?

cbron: If there is a new package without RC I'd assume that is the new production version ? As long as it doesn't also exist in release branch which would be invalid.

aiyengar2:
@StrongMonkey re-requesting your review since I expanded the scope of this PR to take care of several small issues.

templates/template/Makefile
```diff
@@ -1,8 +1,11 @@
-CHARTS_BUILD_SCRIPT_VERSION := fdf0565
```
aiyengar2, line 1: This is tracked directly in a scripts/version file now. That way we won't automatically update the script version on a docs change.

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
 
 {{- end }}
 
+{{- if (eq .Template "source") }}
+### Making Changes
```
aiyengar2, line 83: @nickgerace can you take a look?

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+PACKAGE=<packageName>
+mkdir -p packages/${PACKAGE}
+touch packages/${PACKAGE}/package.yaml
+```
+
+See `packages/README.md` to configure the `packages/${PACKAGE}/package.yaml` file based on the Package that you are planning to add. To make changes, see the steps listed below.
+
+#### If the package already exists
+
+If you are working with a single Package, set `export PACKAGE=<packageName>` to inform the scripts that you only want to make changes to a particular package. This will prevent the scripts from running commands on every package in this repository.
```
nickgerace, line 99: nit: split sentences into multiple lines

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+
+#### If this is the first time you are adding a package:
+
+```shell
+PACKAGE=<packageName>
+mkdir -p packages/${PACKAGE}
+touch packages/${PACKAGE}/package.yaml
+```
+
+See `packages/README.md` to configure the `packages/${PACKAGE}/package.yaml` file based on the Package that you are planning to add. To make changes, see the steps listed below.
```
nickgerace, line 95: nit: split sentences into multiple lines

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+
+#### If the package already exists
+
+If you are working with a single Package, set `export PACKAGE=<packageName>` to inform the scripts that you only want to make changes to a particular package. This will prevent the scripts from running commands on every package in this repository.
+
+You'll also want to update the `packageVersion` and `releaseCandidateVersion` located in `packages/${PACKAGE}/package.yaml`. See the section below for how to update this field.
```
nickgerace, line 101: nit: split sentences into multiple lines

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+make clean # Cleans up your workingDir, leaving behind only the generated-changes/
+```
+
+Once your directory is clean, you are ready to submit a PR.
+
+#### Versioning Packages
+
+If this major.minor.patch version of the Chart has never been released, reset the `packageVersion` to `01` and the `releaseCandidateVersion` to `00`.
+
+If this major.minor.patch version of the Chart has been released before:
```
nickgerace, line 117: nit: wrap major.minor.patch with "\`"  characters like so `major.minor.patch`

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+# Make your changes here to the workingDir directly here
+make patch # Saves changes to generated-changes/
+make clean # Cleans up your workingDir, leaving behind only the generated-changes/
+```
+
+Once your directory is clean, you are ready to submit a PR.
+
+#### Versioning Packages
+
+If this major.minor.patch version of the Chart has never been released, reset the `packageVersion` to `01` and the `releaseCandidateVersion` to `00`.
```
nickgerace, line 115: nit: wrap major.minor.patch with "\`"  characters like so `major.minor.patch`

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+If this major.minor.patch version of the Chart has been released before:
+- If this is the first time you are making a change to this chart for a specific Rancher release (e.g. the current `packageVersion` has already been released in the Live Branch), increment the `packageVersion` by 1 and reset the `releaseCandidateVersion` to `00`.
+- Otherwise, increment the `releaseCandidateVersion` by 1. Leave the `packageVersion` as is.
+
+{{ end -}}
+
+{{- if (eq .Template "staging") }}
+### Cutting a Release
+
+In the Staging branch, cutting a release involves moving the contents of the `assets/` directory into `released/assets/`, deleting the contents of the `charts/` directory, and updating the `index.yaml` to point to the new location for those assets. This process is entirely automated via the `make release` command.
```
nickgerace, line 127: nit: split multiple sentences into new lines

templates/template/README.md
```diff
@@ -79,24 +78,103 @@ Before submitting any PRs, a Github Workflow will check to see if your package d
...
+
+No other changes are expected.
+
+Note: these steps should be taken only after following the steps to cut a release to your Live Branch.
+{{ end -}}
+
+{{- if (eq .Template "live") }}
+### Cutting a Release
+
+In the Live branch, cutting a release requires you to run the `make sync` command. This command will automatically get the latest charts / resources merged into the the branches you sync with (as indicated in this branch's `configuration.yaml`) and will fail if any of those branches try to modify already released assets. If the `make sync` command fails, you might have to manually make changes to the contents of the Staging Branch to resolve any issues.
```
nickgerace, line 144: nit: i'll stop making this comment to avoid spam, but I think creating a newline per new sentence is great for readability and editing.

nickgerace (approved):
All nits, reviewed only `README.md`. LGTM

pkg/sync/compare.go
aiyengar2: Yeah, packages without RC shouldn't exist in the Staging branch in the first place, since it belongs in production. Expanding the scope of the Staging branch to also have packageVersions not scoped to a specific rcVersion isn't something we're tracking right now.

You also would never be able to push something to the Staging Branch that has a version in the Live Branch since the Source Branch validates against both Staging and Live (so it would fail on validating against Live since the packageVersion is already tracked).

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

templates/template/README.md
aiyengar2: addressed in latest, thanks for catching!

aiyengar2:
Double checked that the go templates are rendering correctly by pulling in the docs based on the branch this PR is pointing to at https://github.com/aiyengar2/charts-demo/tree/dev-v2.5-source. This has the latest docs that we would be picking up in each of the Source, Staging, and Live branches.

@nickgerace if you could give it *one* more look when actually rendered, that would be awesome.

aiyengar2:
Added .DS_Store to .gitignore

pkg/sync/compare.go
cbron: Does that mean then for pulling in an upstream non-rancher package at 3.1.0, you have to tag it as 3.1.0-rc1 in staging ? That works but its kinda a misnomer.

aiyengar2: Yes, except it would be 3.1.000-rc1 (where 00 is the packageVersion)

templates/template/Makefile
```diff
@@ -1,8 +1,13 @@
...
-
 pull-scripts:
 	./scripts/pull-scripts
 
+{{ if (eq .Template "staging") -}}
+
+release:
+	./scripts/release-assets
+
+{{ end -}}
```
cbron, line 9: Are you go-templating a makefile ? Guessing this gets pushed out by the go code and then executed inside container.

aiyengar2: Yes, everything within `templates/template` is a Go template based on the `configuration.yaml`. That's how we ship different files to different branches

cbron (approved):

PR #20, opened 2021-03-04
Support doNotReleasePackage

aiyengar2:
Allows a user to specify that a chart should not be released via adding `doNotRelease: true` to the `package.yaml`.

This just adds a flag on top of `make charts`, which in turn blocks `make sync` or `make validate`.

Does not block `make prepare` or `make patch` (if non-local).

The purpose behind this PR is to support avoiding the release of charts like rancher-kiali-server or cilium-original that may not make sense to release as standalone charts in our repository.

### DIFF
--- a/pkg/charts/package.go
+++ b/pkg/charts/package.go
@@ -25,6 +25,8 @@ type Package struct {
 	ReleaseCandidateVersion int `yaml:"releaseCandidateVersion"`
 	// AdditionalCharts are other charts that should be packaged together with this
 	AdditionalCharts []AdditionalChart `yaml:"additionalCharts,omitempty"`
+	// DoNotRelease represents a boolean flag that indicates a package should not be tracked in make charts
+	DoNotRelease bool `yaml:"doNotRelease,omitempty"`
 
 	// fs is a filesystem rooted at the package
 	fs billy.Filesystem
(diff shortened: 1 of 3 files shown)


jcaamano (approved):

pkg/charts/package.go
```diff
@@ -86,6 +88,10 @@ func (p *Package) GeneratePatch() error {
 
 // GenerateCharts creates Helm chart archives for each chart after preparing it
 func (p *Package) GenerateCharts() error {
+	if p.DoNotRelease {
+		logrus.Infof("Skipping package marked doNotRelease")
+		return nil
+	}
```
cbron, line 94: If kiali is the use case here, doesn't istio literally use the kiali tarball ? What I would think accomplishes that would be removing it from index.yaml. 
Maybe there is a better use case where the chart does get packaged in or isn't being used.

jcaamano: The other use case is cilium @ rancher/rke2-charts#46. We want upstream cilium as a package subchart and patched and not to be consumed directly.

aiyengar2: > If kiali is the use case here, doesn't istio literally use the kiali tarball ? 

@cbron at the moment, no. 

Istio [re-declares Kiali](https://github.com/rancher/charts/tree/release-v2.5/charts/rancher-istio/rancher-istio/1.8.300/charts/kiali) within it [as a local chart](https://github.com/rancher/charts/blob/aae93d0125fad561024b25ac1a15580ea612e158/charts/rancher-istio/rancher-istio/1.8.300/requirements.yaml#L4), so they are completely independent of each other.

So we could safely remove it from going through the charts framework at all, unless there is a use case for installing Kiali alone.

cbron: Interesting. No Kiali is strictly for Istio and it is hidden anyway, so we should probably do this.

cbron (approved):

PR #38, opened 2021-05-05
Changes for v0.2.0

aiyengar2:
Highlights:
- Cleaned up code for unused stuff and created some helper functions to reduce redundant code
- Removed `make sync`, `make rebase`
- Renamed `make docs` to `make template`
- Heavily simplified `make validate` and removed a ton of files that were overly complex due to new requirements. See below for more details
- Modified scripts and templates to document the new features
- Updated versioning for semver compliance for Helm 3.5.2
- Added a package-ci script that can be used to validate that charts-build-script changes don't break basic functionality
- Minor bugfixes

Related Issue: https://github.com/rancher/rancher/issues/32199


main.go
```diff
@@ -94,27 +86,31 @@ func main() {
...
-		{
-			Name:   "sync",
-			Usage:  "Pull in new generated assets from branches that the configuration.yaml has set your current branch to sync with",
-			Action: synchronizeRepo,
+			Flags:  []cli.Flag{packageFlag, configFlag},
 		},
 		{
-			Usage:  "Pulls in the latest docs to this repository",
-			Name:   "docs",
-			Action: getDocs,
```
aiyengar2, line 117: Based on feedback from the RKE2 team, I have renamed `make docs` to `make template` since it does way more than just update the docs (e.g. it updates scripts, .gitignore, CNAME, etc. as well).

I also added flags for repositoryUrl and branch to make it easier for us to pull in dev versions of docs.

pkg/charts/package.go
```diff
@@ -21,10 +19,8 @@ type Package struct {
 	Name string `yaml:"name"`
 	// PackageVersion represents the current version of the package. It needs to be incremented whenever there are changes
 	PackageVersion int `yaml:"packageVersion"`
-	// ReleaseCandidateVersion represents the version of the release candidate for a given package.
-	ReleaseCandidateVersion int `yaml:"releaseCandidateVersion"`
 	// AdditionalCharts are other charts that should be packaged together with this
-	AdditionalCharts []AdditionalChart `yaml:"additionalCharts,omitempty"`
+	AdditionalCharts []*AdditionalChart `yaml:"additionalCharts,omitempty"`
```
aiyengar2, line 23: Required to ensure upstreamChartVersion is preserved, since pass-by-value would not preserve this.

pkg/helm/export.go
```diff
@@ -26,7 +27,20 @@ func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, chartVer
...
-	chartVersion = chart.Metadata.Version + chartVersion
+	chartVersionSemver, err := semver.Make(chart.Metadata.Version)
+	if err != nil {
+		return fmt.Errorf("Cannot parse original chart version %s as valid semver", chart.Metadata.Version)
+	}
+	// Add packageVersion as string, preventing errors due to leading 0s
+	if packageVersion >= 999 {
+		return fmt.Errorf("Maximum number of packageVersions is 999, found %d", packageVersion)
+	}
+	chartVersionSemver.Patch = 1000*chartVersionSemver.Patch + uint64(packageVersion)
```
aiyengar2, line 38: I chose 1000 since that gives us the ability to handle up to 999 packageVersion changes and automatically ensures that any of the newly generated charts always have a bigger semver than the old charts unless they have a patchVersion of 0.

pkg/path/path.go
```diff
@@ -1,16 +1,12 @@
 package path
 
 const (
-	// ChartsRepositoryRebasePatchesDir is the directory where a synchronize call will generate any conflicts
-	ChartsRepositoryRebasePatchesDir = "rebase/generated-changes"
-	// ChartsRepositoryCurrentBranchDir is a directory that will be used to store your current assets
-	ChartsRepositoryCurrentBranchDir = "original-assets"
-	// ChartsRepositoryUpstreamBranchDir is a directory that will be used to store the latest copy of a branch you want to sync with
-	ChartsRepositoryUpstreamBranchDir = "new-assets"
```
aiyengar2, line 9: No longer required due to the deprecation of sync and simplification of validate.

pkg/path/path.go
```diff
@@ -1,16 +1,12 @@
...
+	// ChartsRepositoryUpstreamBranchDir is a directory that will be used to store the latest copy of the branch containing your released assets
+	ChartsRepositoryUpstreamBranchDir = "released-assets"
```
aiyengar2, line 5: Used by validate now.

scripts/integration
```diff
@@ -1,87 +0,0 @@
-#!/bin/bash
```
aiyengar2, line 1: Dropped this script entirely in favor of a package-ci script that is far less verbose and run per repository.

It doesn't make sense to run these tests in CI for charts-build-scripts if we are introducing breaking changes.

templates/.github/workflows/push.yaml
```diff
@@ -1,30 +0,0 @@
-name: CI-push
```
aiyengar2, line 1: No more push workflows

templates/live.yaml
```diff
@@ -1,8 +1,4 @@
 template: live
 
-sync:
-- url: https://github.com/rancher/charts.git
-  branch: dev-v2.5
-  dropReleaseCandidates: true
```
aiyengar2, line 6: sync is deprecated. Manual steps have been provided in the README.md

templates/source.yaml
```diff
@@ -1,8 +0,0 @@
-template: source
```
aiyengar2, line 1: No more source branches.

templates/template/scripts/package-ci
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
aiyengar2, line 1: A new script I wrote to test changes as part of this rewrite. This isn't plugged into CI,  but it's a nice-to-have in case we run into issues.

main.go
```diff
@@ -86,15 +84,10 @@ func main() {
 		},
 		{
 			Name:   "validate",
-			Usage:  "Ensure a sync will not overwrite generated assets in branches that the configuration.yaml wants you to validate against",
```
aiyengar2, line 105: `sync` has been entirely removed and `validate` has been heavily modified for the new scripts.

The process for validate is now:
- `make charts`
- Check if Git is clean. If not, fail and **leave the assets, charts, and index.yaml behind**
- `git checkout UPSTREAM -- assets charts` + regenerate index.yaml
- Check if Git is clean. If not, fail and **leave the assets, charts, and index.yaml behind**

The nice part of this approach is that it is far less complex and has the added benefit that simply running `make validate` locally is sufficient to update charts, assets, and index.yaml directory to the latest assets / charts from packages and the released branch!

It also requires only a single command to be run on CI.

aiyengar2:
Adding reviewers for initial checks...

WIP branch to take a look at docs: https://github.com/aiyengar2/charts/tree/bump_staging_v0_2_0

jiaqiluo (changes requested):
@aiyengar2  
Awesome work! I cannot wait to use this new version of charts-build-scripts!

The PR looks good to me overall. I reviewed all changes but did not run the code.

templates/template/packages/README.md
```diff
@@ -82,34 +80,31 @@ Developers will use the following commands to work with packages:
...
 - Run `make clean`
-- Commit your changes and push them to another branch to get ready for a PR
-- Run `make validate` to ensure that your current repository wouldn't introduce any conflicts on a sync. This step can be skipped if a Github Workflow already checks for this on a PR push
-- Open up a PR with your changes
+- Run `make charts` to test your changes by committing the generated directories and pushing it to a branch. Once it is available at such a branch, it can be picked up by any tools that can point to Github based Helm Repositories (e.g. Rancher Cluster Explorer Apps & Marketplace).
+- If you encounter bugs, undo the `make charts` commit (e.g. `git reset HEAD~1 --hard` or `git reset HEAD~1 --soft; git checkout -- .; git clean -df` if you want to do it in multiple steps) and repeat the above steps.
+- Once you are ready to make a PR, ensure the last commit contains only the generated `make charts` changes and submit a PR. The PR worklow will automatically run `make validate` to ensure that your current repository wouldn't introduce any conflicts during a release.
+- Open up a PR with your changes.
 
 ### Troubleshooting
```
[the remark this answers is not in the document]
aiyengar2: addressing in fixup commit.

templates/template/scripts/regenerate-packages
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -e
 
+# Note: These scripts are only intended to migrate from the original build scripts to charts-build-scripts v0.1.x
+# A separate migration process is required for v0.2.x
```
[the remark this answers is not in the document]
aiyengar2: If we need to migrate partner-charts, we'll either need to revisit this script and modify it to be able to perform a migration to v0.2.x or use this script to migrate from bash -> v0.1.x and manually update to v0.2.x. That's why I left it behind as-is

main.go
```diff
@@ -124,202 +120,142 @@ func main() {
...
-	if err != nil {
-		logrus.Fatal(err)
+	_, _, status := getGitInfo()
+	if !status.IsClean() {
+		logrus.Warnf("Git is not clean:\n%s", status)
+		logrus.Fatal("Repository must be clean to run validation")
 	}
+
+	logrus.Infof("Generating charts and checking if Git is clean")
+	CurrentPackage = "" // Validate always runs on all packages
```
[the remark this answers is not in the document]
aiyengar2: I don't think so. CurrentPackage only gets used by `getPackages` in main and this value does not persist across multiple calls to the CLI (e.g. this is always set as a CLI flag on every CLI call), so it should be fine imo.

jiaqiluo (approved):
LGTM

main.go
```diff
@@ -124,202 +120,142 @@ func main() {
...
+		logrus.Fatal("Repository must be clean to generate charts")
 	}
-	packages, err := charts.GetPackages(repoRoot, CurrentPackage)
+	currentBranchRefName, err := repository.GetCurrentBranchRefName(repo)
 	if err != nil {
-		logrus.Fatal(err)
-	}
-	if len(packages) == 0 {
-		logrus.Fatalf("Could not find any packages in packages/")
+		logrus.Warnf("Due to limitations in the Git library used for the scripts, we cannot generate charts in a detached HEAD state.")
```
StrongMonkey, line 148: Nit: looks like there is not parameter so logrus.Warn should be enough

main.go
```diff
@@ -334,18 +270,38 @@ func parseScriptOptions() *options.ChartsScriptOptions {
...
+	repoRoot, err := os.Getwd()
+	if err != nil {
+		logrus.Fatalf("Unable to get current working directory: %s", err)
+	}
+	packages, err := charts.GetPackages(repoRoot, CurrentPackage)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+	if len(packages) == 0 {
+		logrus.Fatalf("Could not find any packages in packages/")
```
StrongMonkey, line 283: same, can use fatal

templates/template/scripts/package-ci
StrongMonkey: Is this intended to run to verify that user manually run through all the steps before they are submitting PRs?

templates/template/packages/README.md
```diff
@@ -82,38 +80,35 @@ Developers will use the following commands to work with packages:
...
-- Repeat above steps if you encounter bugs, otherwise continue.
-- Remove the commit with your newly generated resources (`assets/`, `charts/` and `index.yaml`) and remove the resources themselves manually
 - Run `make clean`
-- Commit your changes and push them to another branch to get ready for a PR
-- Run `make validate` to ensure that your current repository wouldn't introduce any conflicts on a sync. This step can be skipped if a Github Workflow already checks for this on a PR push
-- Open up a PR with your changes
+- Run `make charts` to test your changes by committing the generated directories and pushing it to a branch. Once it is available at such a branch, it can be picked up by any tools that can point to Github based Helm Repositories (e.g. Rancher Cluster Explorer Apps & Marketplace).
+- If you encounter bugs, undo the `make charts` commit (e.g. `git reset HEAD~1 --hard` or `git reset HEAD~1 --soft; git checkout -- .; git clean -df` if you want to do it in multiple steps) and repeat the above steps.
```
cbron, line 105: Or ammend and push force right, nothing special here about undoing commit.

templates/template/scripts/package-ci
cbron: Yea whats point of this exactly.

templates/template/scripts/regenerate-packages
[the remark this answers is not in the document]
cbron: We should delete once they are no longer needed.

pkg/helm/export.go
```diff
@@ -26,7 +27,20 @@ func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, chartVer
 	if err := chart.Validate(); err != nil {
 		return fmt.Errorf("Failed while trying to validate Helm chart: %s", err)
 	}
-	chartVersion = chart.Metadata.Version + chartVersion
+	chartVersionSemver, err := semver.Make(chart.Metadata.Version)
+	if err != nil {
+		return fmt.Errorf("Cannot parse original chart version %s as valid semver", chart.Metadata.Version)
+	}
```
cbron, line 33: Will this fail on the 1.0.000 problem ?

main.go
aiyengar2: updated

main.go
aiyengar2: updated

templates/template/scripts/package-ci
aiyengar2: This isn't really intended to be run by users. It's moreso intended to be run as a sanity check by someone who wants to make a breaking change to the charts-build-scripts and wants to assess whether the repository that is absorbing the changes can be safely migrated (e.g. all commands, except validate, work).

Mostly just built it because I needed the ability to test this, but then checked it in since it seemed like it would be useful in the future.

templates/template/README.md
```diff
@@ -122,53 +99,101 @@ Once your directory is clean, you are ready to submit a PR.
...
-In the Staging branch, cutting a release involves moving the contents of the `assets/` directory into `released/assets/`, deleting the contents of the `charts/` directory, and updating the `index.yaml` to point to the new location for those assets.
-
-This process is entirely automated via the `make release` command.
+If this `major.minor.patch` (e.g. `0.0.1`) version of the Chart has never been released, reset the `packageVersion` to `01`.
+
+If this `major.minor.patch` (e.g. `0.0.1`) version of the Chart has been released before, increment the `packageVersion`.
+
+### Porting over Charts / Assets from another Branch
+
+In the Staging branch, porting over charts from another branch (e.g. `dev-v2.x+1`) requires you to copy the contents of that branch into your Staging branch, which can be done with the following simple Bash script. However, you will need to manually regenerate the Helm index since you only want the index.yaml on the Staging branch to be updated to include the new chart.
```
cbron, line 108: I'm a little confused on this part. Aren't dev and staging the same now ? Meaning isn't dev-v2.5 the staging branch ? The goal of this script I'm guessing is to move across release lines, not entirely clear. 
As for the whole bash script below, can't you just copy paste a folder from one versions branch to another ? Script below is not intuitive, I don't get the point of the forked branch / port.

main.go
cbron: So those bullets listed above are the steps performed when `make validate` is run ?

cbron:
I think the staging/prod branches look good, thanks for doing this. 
I do see another small problem, fleet has 5000 as patch version: https://github.com/aiyengar2/charts/tree/bump_staging_v0_2_0/assets/fleet

templates/template/scripts/regenerate-packages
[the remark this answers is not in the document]
aiyengar2: Yes, we can do this after the partner-chart migration.

templates/template/packages/README.md
aiyengar2: yeah nothing special. I just like to delete the commit entirely though and recreate 😄 .

Should this be changed?

pkg/helm/export.go
StrongMonkey: We might nee to be careful about this. Does it mean the future chart will have something like x.y.z0000? This is probably not going to work well with old charts. Like for example if this is going to apply to old chart let's say 0.3.4, then it will become 0.3.4000 which will become the latest version if we don't update other version because 0.3.4000 is greater than 0.3.500 or 0.3.600. But as long as we are only updating the latest chart we should be good, but sounds like a big version change in our chart today.

pkg/helm/export.go
aiyengar2: If the upstream chart version that was received was itself invalid, then yes I think it would.

templates/template/README.md
aiyengar2: Yeah dev and staging are the same now. The goal of this script is, let's say, moving from `dev-v2.6` to `dev-v2.5`.

You don't want to necessarily maintain two versions of the chart in `packages/`, so you can make the commit to dev-v2.6 with the `packages/` changes and simply port the charts, assets, and index.yaml over to `dev-v2.5`.

Or, if we use @nickgerace 's idea of having a single dev-v2.x branch that contains packages, this would be the normal process to make that version available in `dev-v2.5` or `dev-v2.6` branches (e.g. identical to what we do today with making a commit to `master` and then porting it to `release/v2.5` and `release/v2.6`.

Copying and pasting a folder isn't sufficient since you also need to regenerate the Helm index. But basic idea is the same yeah; you need to copy the charts/, the assets/ and regenerate the Helm index to port over just the generated resources.

You can't do a simple cherry-pick across branches since the index.yaml may have conflicts.

aiyengar2:
> I do see another small problem, fleet has 5000 as patch version:

That's expected right? `5` is the patch and `000` is the packageVersion. Discussing this in https://github.com/rancher/charts-build-scripts/pull/38#discussion_r628659994.

pkg/helm/export.go
aiyengar2: > Does it mean the future chart will have something like x.y.z000?

yes that was my expectation. We could revert this back to a packageVersion with just 2 digits.

re:
> Like for example if this is going to apply to old chart let's say 0.3.4, then it will become 0.3.4000 which will become the latest version if we don't update other version because 0.3.4000 is greater than 0.3.500 or 0.3.600. 

You could just patch the chart version from 0.3.4 => 0.3.0 and increment the packageVersion to 400. Due to the way that the new logic works for versioning (only applies on export), that would result in the right patchVersion you want of 0.3.400.

aiyengar2: @cbron any thoughts? I'm fine to revert back to 2 digit packageVersions.

main.go
aiyengar2: Yes

templates/template/packages/README.md
cbron: Its fine, but I would just say in general be conscious of how much the person has to read. If you said "force push or undo the last commit" you'd have the same affect. Right now they have to read the commands and figure out if something special is going on. For git status thats not a big deal, for longer commands and sequences its quite taxing.

pkg/helm/export.go
cbron: Yea lets do 2 digits, if for nothing else because everyone is familiar with it and there's less edge cases.

templates/template/README.md
cbron: No cherry-pick makes sense, as does index.yaml regeneration. But why not just copy `package/logging` folder, swap branches, and then run `make charts` and `make validate` again ?

aiyengar2: You’ll be committing 2 versions of the logging chart. This is fine, as long as you keep both copies up to date, but if all you want is to make the asset available it makes more sense to have a single source of truth.

aiyengar2 (commented):
> > I do see another small problem, fleet has 5000 as patch version:
> 
> That's expected right? `5` is the patch and `000` is the packageVersion. Discussing this in [#38 (comment)](https://github.com/rancher/charts-build-scripts/pull/38#discussion_r628659994).

Pushed change to switch to 2 digit packageVersions, this should now be 500

templates/template/packages/README.md
aiyengar2: Updated the working for this, made it slightly more verbose but easier to read with no commands.

StrongMonkey (approved):

cbron (approved):
Works for now, still need to figure out how to manage multiple versions in future

Issue #2, opened 2021-01-22
Prevent requiring local builds to pull in charts-build-scripts binary


aiyengar2:
Currently the pull-scripts mechanism manually runs a go build in order to create the binary in the bin/ directory.

We need to switch to a Docker-based approach, which was inhibited before due to Docker slowness associated with making numerous file changes on a mounted volume from the host.

One thing that can enable this is switching the billy.Filesystem to use a memFs instead of an osFs

aiyengar2:
Alternatively, we should at least release the binaries for rancher/charts-build-scripts to avoid having to build the binary using scripts/pull-scripts

PR #42, opened 2021-05-23
Publish binaries & detect version mismatch on pull

aiyengar2:
Also updates Go version to 1.16.

Related Issue: https://github.com/rancher/charts-build-scripts/issues/2


aiyengar2:
@jiaqiluo no rush to review this PR / get it merged in. Would be nice if we can get this in for the 0.2.1 release though.

templates/template/scripts/pull-scripts
```diff
@@ -5,19 +5,37 @@ cd $(dirname $0)
...
 
 rm -rf ../bin
 cd ..
 
-rm -rf charts-build-scripts
-git clone --depth 1 --branch $CHARTS_BUILD_SCRIPT_VERSION $CHARTS_BUILD_SCRIPTS_REPO 2>/dev/null
+mkdir -p bin
+ARCH=$(go version | cut -d' ' -f4 | cut -d'/' -f1)
+curl -s -L ${CHARTS_BUILD_SCRIPTS_REPO%.git}/releases/download/${CHARTS_BUILD_SCRIPT_VERSION}/charts-build-scripts-${ARCH} --output bin/charts-build-scripts
+if [[ $(cat bin/charts-build-scripts) == "Not Found" ]]; then
```
jiaqiluo, line 23: I see the following message when charts-build-scripts does not exist on Mac:

```
> cat bin/charts-build-scripts
cat: bin/charts-build-scripts: No such file or directory
```

jiaqiluo (changes requested):

templates/template/scripts/pull-scripts
aiyengar2: I'm a little confused about what you mean. The previous step curls the binary into `bin/charts-build-scripts`.

Maybe the fix for this is:

```suggestion
if ! [[ -f bin/charts-build-scripts ]] || [[ $(cat bin/charts-build-scripts) == "Not Found" ]]; then 
```

jiaqiluo: Are we comparing the output of the bash command  `cat bin/charts-build-scripts` with the string `Not Found` here? 

In my test, when `bin/charts-build-scripts` does not exist, the output of  `cat bin/charts-build-scripts` command is 
```
cat: bin/charts-build-scripts: No such file or directory
```
not `Not Found`

aiyengar2: Yes, but by definition `bin/charts-build-scripts` should exist if the preceding curl command was successful since we specify `--output bin/charts-build-scripts`.

aiyengar2: I committed the fix suggested in https://github.com/rancher/charts-build-scripts/pull/42#discussion_r641689874, although I didn't seem to need it in local setup; could you try again?

jiaqiluo: Oh, now I understand what it does.  `Not Found` is the output of the curl command and is save as charts-build-scripts 😆

jiaqiluo (approved):
LGTM

aiyengar2:
@cbron this is another PR I drafted for the v0.2.1 release. This one would make all of our CI workflows faster since we no longer have to build the binary locally each time.

cbron (approved):
lgtm, I'm assuming we can't hit our limit of curling github with this.

aiyengar2:
> lgtm, I'm assuming we can't hit our limit of curling github with this.

worst case, it should default to a local build, so we should be fine.

Issue #43, opened 2021-05-24
Remove the target asset and chart before generating charts


jiaqiluo:
Bug.

Reproduce:
delete a file that exists only in the `package/generated-changes/overlay` directory. ie. delete the file that we appended to the chart before. 

Result:
The file still exists in the final chart in the `assets` and `charts` directories. 

Fix:
if the chart already exists in the `assets` and `charts` directories, delete them before exporting the newly generated charts.

jiaqiluo:
no need for QA to test. Close the issue because the PR is merged.

PR #44, opened 2021-05-28
clean the target chart directory if it exists before un-archiving the chart's package

jiaqiluo:
issue: https://github.com/rancher/charts-build-scripts/issues/43

The archived file is updated by the `charts` command, so we only need to clean the chart directory.


aiyengar2 (approved):
LGTM

cbron (approved):

Issue #45, opened 2021-05-28
Add the `version` option to the package.yaml


jiaqiluo:
Instead of deprecating the `packageVersion` option, we will introduce one new option: `version`.

Here are some general rules we will follow:

- rule 1: `version` and `packageVersion` both are optional but can **not** exist at the same time
- rule 2: if the `version` exists, then use it as the final version ( developer needs to make sure ti is a valid semVer)
- rule 3: if the `packageVersion` exists, then add it to the patch version part after the chart's patch files are applied 
- rule 4: if neither `packageVersion` nor `version` exists, we should default to omitting any packageVersion. A great example of where this would be used is fleet, which is a “do not touch me” policy

jiaqiluo:
no need for QA to test. Close the issue because the PR is merged.

PR #48, opened 2021-06-02
add 'version' to the package.yaml file

jiaqiluo:
Issue: https://github.com/rancher/charts-build-scripts/issues/45

The following cases have been tested:
the package.yaml contains
- only version
- only packageVersion
- both version and packageVersion
- none of them

### DIFF
--- a/pkg/charts/package.go
+++ b/pkg/charts/package.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/blang/semver"
 	"github.com/go-git/go-billy/v5"
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
 	"github.com/rancher/charts-build-scripts/pkg/helm"
@@ -17,8 +18,10 @@ type Package struct {
 
 	// Name is the name of the package
 	Name string `yaml:"name"`
+	// Version represents the version of the package. It will override other values if it exists
+	Version *semver.Version `yaml:"version,omitempty"`
 	// PackageVersion represents the current version of the package. It needs to be incremented whenever there are changes
-	PackageVersion int `yaml:"packageVersion"`
+	PackageVersion *int `yaml:"packageVersion"`
 	// AdditionalCharts are other charts that should be packaged together with this
 	AdditionalCharts []*AdditionalChart `yaml:"additionalCharts,omitempty"`
 	// DoNotRelease represents a boolean flag that indicates a package should not be tracked in make charts
@@ -95,12 +98,12 @@ func (p *Package) GenerateCharts() error {
 	packageAssetsDirpath := filepath.Join(path.RepositoryAssetsDir, p.Name)
 	packageChartsDirpath := filepath.Join(path.RepositoryChartsDir, p.Name)
 	// Add PackageVersion to format
-	err := p.Chart.GenerateChart(p.rootFs, p.fs, p.PackageVersion, packageAssetsDirpath, packageChartsDirpath)
+	err := p.Chart.GenerateChart(p.rootFs, p.fs, p.PackageVersion, p.Version, packageAssetsDirpath, packageChartsDirpath)
 	if err != nil {
 		return fmt.Errorf("Encountered error while exporting main chart: %s", err)
 	}
 	for _, additionalChart := range p.AdditionalCharts {
-		err = additionalChart.GenerateChart(p.rootFs, p.fs, p.PackageVersion, packageAssetsDirpath, packageChartsDirpath)
+		err = additionalChart.GenerateChart(p.rootFs, p.fs, p.PackageVersion, p.Version, packageAssetsDirpath, packageChartsDirpath)
 		if err != nil {
 			return fmt.Errorf("Encountered error while exporting %s: %s", additionalChart.WorkingDir, err)
 		}
--- a/pkg/charts/parse.go
+++ b/pkg/charts/parse.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/blang/semver"
 	"github.com/go-git/go-billy/v5"
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
 	"github.com/rancher/charts-build-scripts/pkg/options"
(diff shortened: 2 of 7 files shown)


pkg/options/package.go
```diff
@@ -13,8 +13,10 @@ import (
 // PackageOptions represent the options presented to users to be able to configure the way a package is built using these scripts
 // The YAML that corresponds to these options are stored within packages/<package-name>/package.yaml for each package
 type PackageOptions struct {
+	// Version represents the version of the package. It will override other values if it exists
+	Version string `yaml:"version,omitempty"`
 	// PackageVersion represents the current version of the package. It needs to be incremented whenever there are changes
-	PackageVersion int `yaml:"packageVersion" default:"0"`
+	PackageVersion int `yaml:"packageVersion,omitempty"`
```
aiyengar2, line 19: Why is this change necessary? Wouldn't the default behavior be to set the packageVersion to 0 regardless?

pkg/charts/package.go
```diff
@@ -17,8 +17,10 @@ type Package struct {
 
 	// Name is the name of the package
 	Name string `yaml:"name"`
+	// Version represents the version of the package. It will override other values if it exists
+	Version string `yaml:"version,omitempty"`
 	// PackageVersion represents the current version of the package. It needs to be incremented whenever there are changes
-	PackageVersion int `yaml:"packageVersion"`
+	PackageVersion int `yaml:"packageVersion,omitempty"`
```
aiyengar2, line 23: I think it would make more sense to do the parsing for Semver on trying to parse the PackageOptions. What do you think?

e.g. Let's import `"github.com/blang/semver"` into this file and use `*semver.Version` for the Version instead and change all the signatures for Version to take in a `*semver.Version` as well. A pointer is better since we can check if it is nil.

pkg/charts/parse.go
```diff
@@ -96,6 +100,7 @@ func GetPackage(rootFs billy.Filesystem, name string) (*Package, error) {
 		Chart: chart,
 
 		Name:             name,
+		Version:          packageOpt.Version,
```
aiyengar2, line 103: Use `semver.Make(packageOpt.version)`

pkg/charts/parse.go
```diff
@@ -79,6 +79,10 @@ func GetPackage(rootFs billy.Filesystem, name string) (*Package, error) {
 	if err != nil {
 		return nil, err
 	}
+	// version and packageVersion can not exist at the same time although both are optional
+	if packageOpt.Version != "" && packageOpt.PackageVersion != 0 {
+		return nil, fmt.Errorf("Cannot have both version and packageVersion at the same time")
+	}
```
aiyengar2, line 85: The fact that the `PackageVersion` can no longer be specified as 0 seems like a breaking change to users since this was previously allowed.

Instead of having the `packageOpt.PackageVersion` be an int, can you change it to a pointer to an int? That way, the omitempty logic you have below would also work.

aiyengar2: Wd should also change `version` to be a pointer, since it is also an optional field that can be nil. That way, this statement just checks whether both are non-nil.

pkg/helm/export.go
```diff
@@ -41,13 +41,24 @@ func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, packageV
...
+			return fmt.Errorf("Maximum number of packageVersions is %d, found %d", MaxPatchNum, packageVersion)
+		}
+		chartVersionSemver.Patch = PatchNumMultiplier*chartVersionSemver.Patch + uint64(packageVersion)
+	} else if version != "" {
+		chartVersionSemver, err = semver.Parse(version)
+		if err != nil {
+			return fmt.Errorf("Cannot parse version %s as an valid semver: %s", version, err)
+		}
+	} else {
+		logrus.Infof("No version or packageVersion is provided in package.yaml, using the chart version with patch files applied")
```
aiyengar2, line 56: I think we can drop this log

pkg/helm/export.go
```diff
@@ -41,13 +41,24 @@ func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, packageV
...
+		// Add packageVersion as string, preventing errors due to leading 0s
+		if uint64(packageVersion) >= MaxPatchNum {
+			return fmt.Errorf("Maximum number of packageVersions is %d, found %d", MaxPatchNum, packageVersion)
+		}
+		chartVersionSemver.Patch = PatchNumMultiplier*chartVersionSemver.Patch + uint64(packageVersion)
+	} else if version != "" {
+		chartVersionSemver, err = semver.Parse(version)
+		if err != nil {
+			return fmt.Errorf("Cannot parse version %s as an valid semver: %s", version, err)
+		}
```
aiyengar2, line 54: We shouldn't even try parsing `semver.Make(chart.Metadata.Version)` if the version is specified. Can you move that call into the packageVersion check?

We should probably restructure this as:

1. If version, take it.
2. Else, parse the chart.Metadata.Version. If packageVersion, append it.

pkg/helm/export.go
```diff
@@ -41,13 +41,24 @@ func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, packageV
...
+	} else {
+		logrus.Infof("No version or packageVersion is provided in package.yaml, using the chart version with patch files applied")
 	}
-	chartVersionSemver.Patch = PatchNumMultiplier*chartVersionSemver.Patch + uint64(packageVersion)
-	if len(upstreamChartVersion) > 0 {
+
+	if len(upstreamChartVersion) > 0 && upstreamChartVersion != chartVersionSemver.String() {
 		// Add buildMetadataFlag for forked charts
+		logrus.Infof("The chart's version is different from the upstream chart's, appending the upstream's version to its meatadata")
```
aiyengar2, line 61: Don't think this log statement is necessary

aiyengar2 (changes requested):

pkg/options/package.go
jiaqiluo: The change is discarded.

aiyengar2 (approved):
LGTM

pkg/charts/parse.go
```diff
@@ -79,6 +80,19 @@ func GetPackage(rootFs billy.Filesystem, name string) (*Package, error) {
 	if err != nil {
 		return nil, err
 	}
+	// version and packageVersion can not exist at the same time although both are optional
+	if packageOpt.Version != nil && packageOpt.PackageVersion != nil {
+		return nil, fmt.Errorf("Cannot have both version and packageVersion at the same time")
+	}
+	var version *semver.Version
+	if packageOpt.Version != nil {
+		temp, err := semver.Make(*packageOpt.Version)
```
cbron, line 89: Might run this through Matt's validator as well: https://github.com/mattfarina/semver-isvalid/blob/main/pkg/semver/valid.go

cbron (approved):
lgtm, comment is up for discussion.

pkg/charts/parse.go
jiaqiluo: `semver.Make` function also does the validation as a part of the making process. 
I skim through both implementations and do not see many differences in the items/rules they check.  

We can use Matt's validator for sure if we want.

PR #55, opened 2021-07-16
update the README for packages

jiaqiluo:
- [ ] After the PR is merged, run `make template` in the dev branches in rancher/charts to update templates.

### DIFF
--- a/templates/template/README.md
+++ b/templates/template/README.md
@@ -1,5 +1,5 @@
 {{- if (eq .Template "staging") -}}
-## Staging Branch
+## Dev Branch
 
 This branch contains Packages and generated assets that have not been officially released yet.
 
@@ -7,7 +7,7 @@ See the README.md under `packages/`, `assets/`, and `charts/` for more informati
 
 The following directory structure is expected:
 ```text
-package/
+packages/
   <package>/
   ...
 assets/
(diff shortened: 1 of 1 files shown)


templates/template/README.md
```diff
@@ -21,6 +21,12 @@ charts/
         # Unarchived Helm chart
   ...
 ```
+
+In some cases, the `package` of a chart does not exist in all dev branches:
+- If the chart is released in rancher v2.5 originally and no longer developed in rancher v2.6, its package will be reserved in the dev-v2.5 branch, removed from the dev-v2.6 branch. An example will be the package for Rio.
+- If a new version of the chart is released in v2.5, we will forward port the new chart to dev-v2.6 and release-v2.6 branches. See the section `Porting over Charts / Assets from another Branch` below for instructions.
+- if we need to develop the chart in 2.6 again, we will bring its package back if it was removed in dev-v2.6 in the past.
```
aiyengar2, line 28: Since these scripts are used by partner-charts, Opni, etc, we should not add in Rancher-specific language

```suggestion

### "Orphaned" Charts

Not all charts tracked in `charts/` and `assets/` will have a corresponding entry in `packages/`, but all charts tracked in `packages/` must exist in `charts/` and `assets/` to pass a `make validate`.

Some situations that would merit "orphaned" charts include, but or not limited to:
- Charts that are no longer in active development (e.g. deprecated charts, older versions of released charts)
- Charts that are forward-ported / back-ported over from other Helm repositories

Note: as long as it doesn't modify an already released version of a chart (which will be checked on a `make validate`), you can always bring back a previously removed package to the `packages/` directory to move it back to active development.
```

aiyengar2 (changes requested):

nickgerace:
@jiaqiluo will review after Arvind's approval due to large comment.

aiyengar2 (approved):
LGTM

nickgerace (approved):

PR #58, opened 2021-08-19
Output error when package is not set for patch

aiyengar2:
It's never reasonable to run a patch on **everything** in the repository. So this commit ensures that you can only run patch when a specific package is set.

```
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts
$ make patch
FATA[0000] No PACKAGE is set. Run export PACKAGE=<package> before running this command.
make: *** [patch] Error 1
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -130,10 +130,11 @@ func prepareCharts(c *cli.Context) {
 
 func generatePatch(c *cli.Context) {
 	packages := getPackages()
-	for _, p := range packages {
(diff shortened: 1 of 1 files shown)


aiyengar2:
Not a high priority, but a nice-to-have if we can get reviews!

jiaqiluo (approved):
LGTM

kinarashah (approved):

cbron (approved):

PR #64, opened 2021-10-19
add the step of combining all CRD files into a single compressed file

jiaqiluo:
Issue:
In the CRD chart, Helm combines all CRD files into one file and saves the file into a configMap, then Helm runs a pod to install/uninstall CRDs by reading the file from the configMap. 
But in the latest monitoring v2 chart, the creation of the configMap fails because the size exceeds the 1MB limit set by k8s. 
Also, the issue cannot be fixed in the Helm chart since Helm does not support compressing files. 

Fix:
When preparing the CRD chart, a new file that combines all CRDs is created and compressed by gzip. Then Helm chart can use this file to create the configMap. 
The CRD-chart templates in the rancher-monitoring package also need to be updated in order to use this new file. https://github.com/rancher/charts/pull/1567

### DIFF
--- a/pkg/charts/additionalchart.go
+++ b/pkg/charts/additionalchart.go
@@ -44,16 +44,22 @@ func (c *AdditionalChart) ApplyMainChanges(pkgFs billy.Filesystem) error {
 	if err != nil {
 		return fmt.Errorf("Encountered error while trying to get the main chart's working directory: %s", err)
 	}
-	if err := helm.CopyCRDsFromChart(pkgFs, mainChartWorkingDir, path.ChartCRDDir, c.WorkingDir, c.CRDChartOptions.CRDDirectory); err != nil {
-		return fmt.Errorf("Encountered error while trying to copy CRDs from %s to %s: %s", mainChartWorkingDir, c.WorkingDir, err)
+	if c.CRDChartOptions.UseTarArchive {
+		if err := helm.ArchiveCRDs(pkgFs, mainChartWorkingDir, path.ChartCRDDir, c.WorkingDir, c.CRDChartOptions.CRDDirectory); err != nil {
+			return fmt.Errorf("encountered error while trying to bundle and compress CRD files from the main chart: %s", err)
+		}
+	} else {
+		if err := helm.CopyCRDsFromChart(pkgFs, mainChartWorkingDir, path.ChartCRDDir, c.WorkingDir, c.CRDChartOptions.CRDDirectory); err != nil {
+			return fmt.Errorf("Encountered error while trying to copy CRDs from %s to %s: %s", mainChartWorkingDir, c.WorkingDir, err)
+		}
+	}
+	if c.CRDChartOptions.AddCRDValidationToMainChart {
+		if err := AddCRDValidationToChart(pkgFs, mainChartWorkingDir, mainChartWorkingDir, path.ChartCRDDir); err != nil {
+			return fmt.Errorf("Encountered error while trying to add CRD validation to %s based on CRDs in %s: %s", mainChartWorkingDir, c.WorkingDir, err)
+		}
 	}
 	if err := helm.DeleteCRDsFromChart(pkgFs, mainChartWorkingDir); err != nil {
 		return fmt.Errorf("Encountered error while trying to delete CRDs from main chart: %s", err)
-	}
-	if c.CRDChartOptions.AddCRDValidationToMainChart {
-		if err := AddCRDValidationToChart(pkgFs, mainChartWorkingDir, c.WorkingDir, c.CRDChartOptions.CRDDirectory); err != nil {
-			return fmt.Errorf("Encountered error while trying to add CRD validation to %s based on CRDs in %s: %s", mainChartWorkingDir, c.WorkingDir, err)
-		}
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


pkg/charts/additionalchart.go
```diff
@@ -47,6 +47,9 @@ func (c *AdditionalChart) ApplyMainChanges(pkgFs billy.Filesystem) error {
 	if err := helm.CopyCRDsFromChart(pkgFs, mainChartWorkingDir, path.ChartCRDDir, c.WorkingDir, c.CRDChartOptions.CRDDirectory); err != nil {
 		return fmt.Errorf("Encountered error while trying to copy CRDs from %s to %s: %s", mainChartWorkingDir, c.WorkingDir, err)
 	}
+	if err := filesystem.ArchiveDir(pkgFs, c.WorkingDir, c.CRDChartOptions.CRDDirectory, filepath.Join(c.WorkingDir, path.ChartExtraFileDir)); err != nil {
```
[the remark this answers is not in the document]
jiaqiluo: > Either we should ArchiveDir or CopyCRDsFromChart, is there any reason why a CRD chart should need both?

`CopyCRDsFromChart` is kept because the rancher-monitoring-crd chart needs those crd files to render its final charts. Keeping them makes the chance on both here and the monitoring chart relatively easy and small.

pkg/filesystem/filesystem.go
```diff
@@ -249,6 +249,52 @@ func UnarchiveTgz(fs billy.Filesystem, tgzPath, tgzSubdirectory, destPath string
 	return nil
 }
 
+// ArchiveDir archives a directory or a file into a tgz file and put it at destTgzPath which should end with .tgz
+func ArchiveDir(fs billy.Filesystem, srcPath, destTgzPath string) error {
+	if !strings.HasSuffix(destTgzPath, ".tgz") {
+		return fmt.Errorf("the destTgzPath %s does not end with .tgz", destTgzPath)
```
aiyengar2, line 255:
```suggestion
		return fmt.Errorf("cannot archive %s to %s since the archive path does not end with '.tgz'", srcPath, destTgzPath)
```

pkg/filesystem/filesystem.go
```diff
@@ -249,6 +249,52 @@ func UnarchiveTgz(fs billy.Filesystem, tgzPath, tgzSubdirectory, destPath string
...
+		header, err := tar.FileInfoHeader(info, info.Name())
+		if err != nil {
+			return err
+		}
+		// overwrite the name to be the full path to the file
+		header.Name = path
+		if err := tarWriter.WriteHeader(header); err != nil {
+			return err
+		}
+		// The directory structure is preserved, but there is no data to read from a Dir
```
aiyengar2, line 283:
```suggestion
		// The directory structure is preserved, but there is no data to read from a directory
```

pkg/filesystem/filesystem.go
```diff
@@ -150,7 +150,7 @@ func CopyFile(fs billy.Filesystem, srcPath string, dstPath string) error {
 	return nil
 }
 
-// GetChartArchive gets a chart tgz file from a url and drops it into the path specified on the filesystem
+// GetChartArchive gets a chart tgz file from an url and drops it into the path specified on the filesystem
```
aiyengar2, line 153: should remain `a url`, not `an url`. iirc it's a weird english rule because you say it as `YOU-ARE-EL`, which means it starts with a consonant sound so `a` should be used instead of `an`.

v.s. `an undertaking` is correct since it starts with the traditional `u` sound.

English is weird 😅

pkg/charts/additionalchart.go
```diff
@@ -66,12 +71,14 @@ func (c *AdditionalChart) RevertMainChanges(pkgFs billy.Filesystem) error {
...
 	}
 	if c.CRDChartOptions == nil {
+		// return if the additional chart is not a CRD chart
 		return nil
 	}
 	mainChartWorkingDir, err := c.getMainChartWorkingDir(pkgFs)
 	if err != nil {
 		return fmt.Errorf("Encountered error while trying to get the main chart's working directory: %s", err)
 	}
+	// copy CRD files from packages/<package>/charts-crd/crd-manifest/ to packages/<package>/charts/crds/
```
aiyengar2, line 81: Should we put it in `packages/<package>/charts/crds/` or `packages/<package>/charts/files/`? I'd prefer the latter since I think the `crd` directory is a special directory in Helm, so I'm worried that adding something there could create a future condition that might break... v.s. putting arbitrary files in `files/` is generally a standard practice iirc.

aiyengar2 (changes requested):
wooo great work! nit comments but overall looks great to me.

almost ready for an approval!

pkg/filesystem/filesystem.go
jiaqiluo: fixed.

pkg/filesystem/filesystem.go
jiaqiluo: fixed.

pkg/options/additionalchart.go
```diff
@@ -18,4 +18,6 @@ type CRDChartOptions struct {
 	CRDDirectory string `yaml:"crdDirectory" default:"templates"`
 	// Whether to add a validation file to your main chart to check that CRDs exist
 	AddCRDValidationToMainChart bool `yaml:"addCRDValidationToMainChart"`
+	// Whether to bundle and compress CRD files into a tgz file
+	UseTarArchive bool `yaml:"useTarArchive"`
```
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/helm/crds.go
```diff
@@ -49,3 +49,17 @@ func DeleteCRDsFromChart(fs billy.Filesystem, helmChartPath string) error {
 	}
 	return nil
 }
+
+// ArchiveCRDs bundles, compress and save the CRD files from the source to the destination
```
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/filesystem/filesystem.go
jiaqiluo: fixed.

aiyengar2 (approved):
LGTM

pkg/filesystem/filesystem.go
```diff
@@ -249,6 +249,52 @@ func UnarchiveTgz(fs billy.Filesystem, tgzPath, tgzSubdirectory, destPath string
 	return nil
 }
 
+// ArchiveDir archives a directory or a file into a tgz file and put it at destTgzPath which should end with .tgz
```
cbron, line 252: Surprised there wasn't something in standard library for this.

pkg/filesystem/filesystem.go
```diff
@@ -249,6 +249,52 @@ func UnarchiveTgz(fs billy.Filesystem, tgzPath, tgzSubdirectory, destPath string
...
+	}
+	defer tgzFile.Close()
+
+	gz := gzip.NewWriter(tgzFile)
+	defer gz.Close()
+
+	tarWriter := tar.NewWriter(gz)
+	defer tarWriter.Close()
+
+	return WalkDir(fs, srcPath, func(fs billy.Filesystem, path string, isDir bool) error {
```
cbron, line 269: nit: This indentation is slightly confusing

cbron (approved):
Glossed over it, seems fine to me. Proof will be in how well it works.

PR #66, opened 2021-11-16
fix the validation process 

jiaqiluo:
Add the support for the crd tgz file within the crd chart during the validation process. The crd tgz file is always marked as `Modified` because its metadata is changed when re-generating the chart.  If the crd tgz file is the only files changes in the crd chart, we should ignore it; Otherwise, keep all changes in the crd chart. 

Note:
`make validation` passed when building the binary and running it again my working branch in rancher/charts repo locally

### DIFF
--- a/main.go
+++ b/main.go
@@ -160,6 +160,7 @@ func generateCharts(c *cli.Context) {
 	// Copy in only assets from charts that have changed
 	_, wt, status := getGitInfo()
 	modifiedAssets := make(map[string]bool)
+	modifiedCRDArchives := make(map[string]bool)
 	for p := range status {
 		if p == path.RepositoryHelmIndexFile {
 			wt.Excludes = append(wt.Excludes, gitignore.ParsePattern(p, []string{}))
@@ -182,7 +183,15 @@ func generateCharts(c *cli.Context) {
 			// New chart was introduced
 			modifiedAssetPath = fmt.Sprintf("%s/*", filepath.Join(path.RepositoryAssetsDir, splitPath[1], splitPath[2]))
 		default:
-			// Existing chart was modified
+			// e.g. p = charts/rancher-monitoring/rancher-monitoring-crd/100.1.0+up19.0.3/files/crd-manifest.tgz
+			if strings.HasSuffix(splitPath[2], "-crd") {
+				if strings.HasSuffix(p, ".tgz") {
+					continue
+				}
+				chartKey := filepath.Join(splitPath[0], splitPath[1], splitPath[2], splitPath[3], path.ChartExtraFileDir, path.ChartCRDTgzFilename)
+				modifiedCRDArchives[chartKey] = true
+			}
+			// Existing chart (excluding the crd tgz file) was modified
 			modifiedAssetPath = fmt.Sprintf("%s-%s.tgz", filepath.Join(path.RepositoryAssetsDir, splitPath[1], splitPath[2]), splitPath[3])
 		}
 		// Add chart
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -182,7 +184,16 @@ func generateCharts(c *cli.Context) {
...
-			// Existing chart was modified
+			if strings.HasSuffix(splitPath[2], "-crd") {
+				// e.g. p = charts/rancher-monitoring/rancher-monitoring-crd/100.1.0+up19.0.3/files/crd-manifest.tgz
+				chartKey := fmt.Sprintf("%s-%s-%s", splitPath[1], splitPath[2], splitPath[3])
+				if strings.HasSuffix(p, ".tgz") {
+					modifiedCrdTgzPaths[chartKey] = p
+					continue
+				}
+				modifiedCrdNormalFiles[chartKey] = true
+			}
```
aiyengar2, line 195: I think we should just skip if the CRD tgz file changes, since we know the tgz file will always change... And if an underlying file was modified, it would have added the CRD chart to the loop anyways.

Why can't we just modify all this logic to:

```go
if strings.HasSuffix(p, ".tgz") {
	// e.g. p = charts/rancher-monitoring/rancher-monitoring-crd/100.1.0+up19.0.3/files/crd-manifest.tgz
	continue
}
```

main.go
```diff
@@ -196,6 +207,14 @@ func generateCharts(c *cli.Context) {
...
 		logrus.Infof("Outputted %s", modifiedAssetPath)
 	}
+	for chartKey := range modifiedCrdNormalFiles {
+		if path, found := modifiedCrdTgzPaths[chartKey]; found {
+			// add the crd tgz file within the crd chart
+			wt.Excludes = append(wt.Excludes, gitignore.ParsePattern(path, []string{}))
+			logrus.Infof("Outputted %s", path)
+		}
+	}
+
```
aiyengar2, line 217: can omit all of this with the above suggestion

aiyengar2 (changes requested):
All we need to do here is avoid using `.tgz` files as the basis for adding an asset to `wt.Excludes` (which represents the list of assets that had actual changes) right?

main.go
aiyengar2: Talked to Jack offline; the issue here is that we need to add the underlying `crd-manifest.tgz` into the `wt.Excludes`, and we only want to add it once. This is why an intermediary map is required,

main.go
```diff
@@ -182,7 +184,16 @@ func generateCharts(c *cli.Context) {
...
 			modifiedAssetPath = fmt.Sprintf("%s/*", filepath.Join(path.RepositoryAssetsDir, splitPath[1], splitPath[2]))
 		default:
-			// Existing chart was modified
+			if strings.HasSuffix(splitPath[2], "-crd") {
+				// e.g. p = charts/rancher-monitoring/rancher-monitoring-crd/100.1.0+up19.0.3/files/crd-manifest.tgz
+				chartKey := fmt.Sprintf("%s-%s-%s", splitPath[1], splitPath[2], splitPath[3])
+				if strings.HasSuffix(p, ".tgz") {
+					modifiedCrdTgzPaths[chartKey] = p
+					continue
+				}
```
aiyengar2, line 193: As discussed offline, let's see if we can eliminate this second map in favor of just having the key be something like:
```go
chartKey := filepath.Join(splitPath[0], splitPath[1], splitPath[2]), splitPath[3], "files", "crd-manifest.tgz")
```

aiyengar2 (changes requested):

main.go
jiaqiluo: This is fixed. 
I also move the tgz filename to a const variable.

aiyengar2 (approved):
LGTM; the edge cases around CRD charts in this PR also need to be addressed in https://github.com/rancher/charts-build-scripts/pull/65.

PR #69, opened 2022-01-05
Support nested tgz and tar comparisions

aiyengar2:
Due to the changes introduced in https://github.com/rancher/charts-build-scripts/pull/64, we now have tgz files within tgz files, which was previously unexpected. Therefore, when CompareTgz (used to see whether an archive is up to date by deep comparing the contents) encounters a tgz or tar file, we need to recursively apply the same logic to compare the internal tgz or tar file.

### DIFF
--- a/pkg/filesystem/filesystem.go
+++ b/pkg/filesystem/filesystem.go
@@ -431,10 +431,10 @@ func compareTars(leftFile, rightFile io.Reader) (bool, error) {
 		// Check if both archives contain the files
 		switch {
 		case len(hashes[0]) == 0:
-			logrus.Debugf("%s only exists in left tar", filename)
+			logrus.Debugf("%s does not exist in left tar", filename)
 			identical = false
 		case len(hashes[1]) == 0:
-			logrus.Debugf("%s only exists in right tar", filename)
+			logrus.Debugf("%s does not exist in right tar", filename)
 			identical = false
 		case hashes[0] != hashes[1]:
 			// Hashes do not match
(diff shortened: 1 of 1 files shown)


pkg/filesystem/filesystem.go
```diff
@@ -304,48 +314,114 @@ func CompareTgzs(fs billy.Filesystem, leftTgzPath string, rightTgzPath string) (
...
+				var b bytes.Buffer
+				if _, err := b.ReadFrom(leftTarReader); err != nil {
+					return false, fmt.Errorf("could not read contents of %s: %v", left.Name, err)
+				}
+				r, ok := tgzMap[left.Name]
+				if !ok {
+					r = make([]io.Reader, 2)
+				}
+				r[0] = &b
+				tgzMap[left.Name] = r
```
aiyengar2, line 335: Mirrors the same logic we see below with the hashMap, except the map and the value we want to add to the map (e.g. an io.Reader) is different.

Why can't we just compare tgzs or tars here?
-> Because it's not guaranteed that leftTarReader or rightTarReader are reading files in the same order. This is why we collect it into a map whose keys are the filename and whose values are [leftContents, rightContents] and later run the comparison once the values have both been set.

PennyScissors (approved):

pkg/filesystem/filesystem.go
```diff
@@ -431,10 +431,10 @@ func compareTars(leftFile, rightFile io.Reader) (bool, error) {
 		// Check if both archives contain the files
 		switch {
 		case len(hashes[0]) == 0:
-			logrus.Debugf("%s only exists in left tar", filename)
+			logrus.Debugf("%s does not exist in left tar", filename)
```
prachidamle, line 434: Was this log statement incorrect previously? Don't see any new logic changes around it

aiyengar2: Before it said "the file **only** exists in **right** tar", whereas now it says "the file **no longer** exists in **left** tar"

Since we switched both the side and the meaning, it's the same log

prachidamle (approved):
LGTM

PR #72, opened 2022-01-10
Fix error message on modifying versions of existing charts and add additional docs

aiyengar2:
Validation should not pass if a chart that exists in upstream is removed **and** an entry does not exist in the release.yaml for it.

Since this use case will occur often for pre-release charts (e.g. on transitioning `0.1.2-rc3 -> 0.1.2-rc4`), I added additional docs on it as well for what needs to be done.

### DIFF
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -29,7 +29,7 @@ type CompareGeneratedAssetsResponse struct {
 
 // PassedValidation returns whether the response seems to indicate that the chart repositories are in sync
 func (r CompareGeneratedAssetsResponse) PassedValidation() bool {
-	return len(r.ModifiedPostRelease) == 0 && len(r.UntrackedInRelease) == 0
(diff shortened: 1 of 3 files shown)


pkg/validate/validate.go
```diff
@@ -29,7 +29,7 @@ type CompareGeneratedAssetsResponse struct {
 
 // PassedValidation returns whether the response seems to indicate that the chart repositories are in sync
 func (r CompareGeneratedAssetsResponse) PassedValidation() bool {
-	return len(r.ModifiedPostRelease) == 0 && len(r.UntrackedInRelease) == 0
+	return len(r.ModifiedPostRelease) == 0 && len(r.UntrackedInRelease) == 0 && len(r.UntrackedInRelease) == 0
```
PennyScissors, line 32: Did you intentionally add this duplicated conditional?

aiyengar2: yes 😓, thank you lol

aiyengar2: updated

PennyScissors (approved):

brendarearden (approved):
lgtm

PR #78, opened 2022-02-16
Fix broken validation of charts that exists in both upstream and local

aiyengar2:


pkg/validate/validate.go
```diff
@@ -135,7 +135,7 @@ func CompareGeneratedAssets(repoFs billy.Filesystem, u options.UpstreamOptions,
 		return copyAndUnzip(repoFs, upstreamPath, localPath)
 	}
 
-	localAndUpstream := func(fs billy.Filesystem, upstreamPath, localPath string, isDir bool) error {
+	localAndUpstream := func(fs billy.Filesystem, localPath, upstreamPath string, isDir bool) error {
```
aiyengar2, line 138: the signature of the both function is supposed to have left first and right second: https://github.com/rancher/charts-build-scripts/blob/bc4c3dadee9e35ddf52a6bf4e95e319642097dd1/pkg/filesystem/filesystem.go#L638

Here, while our left is local and our right is upstream (as seen in the following snippet), we flipped it in the function signature.

https://github.com/rancher/charts-build-scripts/blob/bc4c3dadee9e35ddf52a6bf4e95e319642097dd1/pkg/validate/validate.go#L173-L175

Therefore, no checks were actually performed when a chart exists in local and upstream till now.

PennyScissors (approved):

luthermonson (approved):

PR #80, opened 2022-04-28
Refactor main to create mage targets

dbason:
This pull request refactors the CLI functions into exported functions that can be used as mage targets.  The original functions wrap the new ones with a CLI context.

The validate function has been split into 3 separate functions to make it easier to use the remote or local options in mage.

### DIFF
--- a/main.go
+++ b/main.go
@@ -221,12 +221,16 @@ func main() {
 }
 
 func listPackages(c *cli.Context) {
+	List(PorcelainMode)
+}
+
+func List(porcelainMode bool) {
 	repoRoot := getRepoRoot()
 	packageList, err := charts.ListPackages(repoRoot, CurrentPackage)
 	if err != nil {
 		logrus.Fatal(err)
 	}
-	if PorcelainMode {
+	if porcelainMode {
 		fmt.Println(strings.Join(packageList, " "))
 		return
 
@@ -235,7 +239,11 @@ func listPackages(c *cli.Context) {
 }
 
 func prepareCharts(c *cli.Context) {
-	packages := getPackages()
+	Prepare(CurrentPackage)
+}
+
+func Prepare(currentPackage string) {
+	packages := getPackages(currentPackage)
 	if len(packages) == 0 {
 		logrus.Fatal("Could not find any packages in packages/")
 	}
@@ -247,7 +255,11 @@ func prepareCharts(c *cli.Context) {
 }
 
 func generatePatch(c *cli.Context) {
-	packages := getPackages()
+	Patch(CurrentPackage)
+}
+
+func Patch(currentPackage string) {
+	packages := getPackages(currentPackage)
 	if len(packages) == 0 {
 		logrus.Infof("No packages found.")
 		return
@@ -259,7 +271,7 @@ func generatePatch(c *cli.Context) {
 		}
 		logrus.Fatalf(
 			"PACKAGE=\"%s\" must be set to point to exactly one package. Currently found the following packages: %s",
-			CurrentPackage, packageNames,
+			currentPackage, packageNames,
 		)
 	}
 	if err := packages[0].GeneratePatch(); err != nil {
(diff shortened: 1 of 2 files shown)


kralicky:
I believe the exported functions will need to be moved to a different package, since package main can't be imported.

aiyengar2:
Why do we need the functions to be exposed as mage targets?

main.go
```diff
@@ -221,194 +211,64 @@ func main() {
...
-	logrus.Info("Zipping charts to ensure that contents of assets, charts, and index.yaml are in sync.")
-	zipCharts(c)
-
-	logrus.Info("Doing a final check to ensure Git is clean")
-	_, _, status = getGitInfo()
-	if !status.IsClean() {
-		logrus.Warnf("Git is not clean:\n%s", status)
-		logrus.Fatal("Repository must be clean to pass validation")
+		actions.ValidateRemote()
+		return
```
aiyengar2, line 259: there is already a Validate function that we should call in the actions repo, otherwise it should be removed from there.

main.go
```diff
@@ -441,29 +289,11 @@ func getRepoRoot() string {
 	return repoRoot
 }
 
-func getPackages() []*charts.Package {
+func getPackages(currentPackage string) []*charts.Package {
```
aiyengar2, line 292: function should be removed? unused.

aiyengar2: same with getRepoRoot if this is not used. Both are in action package now.

pkg/actions/actions.go
```diff
@@ -0,0 +1,291 @@
...
+	checkGit()
+	logrus.Infof("Successfully validated that current charts and assets are up to date.")
+
+	logrus.Infof("Running local validation only, skipping pulling upstream")
+
+	logrus.Info("Zipping charts to ensure that contents of assets, charts, and index.yaml are in sync.")
+	Zip("")
+
+	logrus.Info("Doing a final check to ensure Git is clean")
+	checkGit()
```
aiyengar2, line 165: logs should be removed

pkg/actions/actions.go
```diff
@@ -0,0 +1,291 @@
...
+
+	pullUpstream(chartsScriptOptions)
+
+	logrus.Info("Zipping charts to ensure that contents of assets, charts, and index.yaml are in sync.")
+	Zip("")
+
+	logrus.Info("Doing a final check to ensure Git is clean")
+	checkGit()
+
+	logrus.Info("Successfully validated current repository!")
```
aiyengar2, line 145: I think this should just call ValidateLocal and ValidateRemote and we should modify the log messages accordingly to make it look good.

I also think this should take in `remote` and `local` bools as function arguments and do the checking to see that exactly 1 of them evaluates to true within this function rather than in the main.go.

pkg/actions/actions.go
```diff
@@ -0,0 +1,291 @@
...
+	return &chartsScriptOptions
+}
+
+func checkGit() {
+	_, _, status := getGitInfo()
+	if !status.IsClean() {
+		logrus.Warnf("Git is not clean:\n%s", status)
+		logrus.Fatal("Repository must be clean to run validation")
+	}
+}
```
aiyengar2, line 242: I think this should be modified to `checkIfGitClean` or something more indicative of what we are checking Git for. I also think we should log the `Git is clean` log here itself.

aiyengar2 (changes requested):
Could the functions in the Action package return errors or do we intentionally want them to not emit errors to allow it to satisfy a function signature mage expects? If we can have them return the log Fatals as errors instead, that would be better.

Also many comments, but thanks for the PR!

pkg/actions/actions.go
dbason: In the current flow if you call it with both local and remote false the logic is to do both local and remove checking.  This makes sense with a CLI, however with mage if you call `mage validate false false` to me it would seem somewhat counterintuitive.  That's why I split them into multiple functions so a mage user could call `mage validatelocal` or `mage validateremote` to call remote or local only.

One better way might be to have constant string types e.g.
```
type ValidateType string
const (
  ValidateAll ValidateType = ""
  ValidateLocal ValidateType = "local"
  ValidateRemote ValidateType = "remote"
)
```
Because of the restrictions in mage the validate method would need to accept a string argument, but this might be a cleaner way to deal with it?

PR #84, opened 2022-07-15
Add rancher img validation

superseb:
First attempt at checking if all images start with `rancher/`. This is to catch this issue as early as possible, most of the code was taken from `rancher/rancher` as the test also happens there. See `pkg/image` for the code in `rancher/rancher`.

I also added a `list-asset-images` command that I want to use to get the images from added assets in a PR and check if the images actually exist in DockerHub to avoid issues later in the Rancher release.

Let me know if this needs to be done differently, or maybe added somewhere else.

### DIFF
--- a/main.go
+++ b/main.go
@@ -122,6 +122,12 @@ func main() {
 			Flags:  []cli.Flag{packageFlag, porcelainFlag},
 		},
 		{
+			Name:   "list-asset-images",
+			Usage:  "Print a list of all images in an asset archive",
+			Action: listAssetImages,
+			Flags:  []cli.Flag{assetFlag, porcelainFlag},
+		},
+		{
 			Name:   "prepare",
 			Usage:  "Pull in the chart specified from upstream to the charts directory and apply any patch files",
 			Action: prepareCharts,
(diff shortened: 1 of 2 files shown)


superseb:
Yeah thought about that but it would mean a change in Rancher code as its not identical code currently, this also means we are blocked by Rancher to make changes. Also I think we will pull in too many dependencies if we depend on Rancher.

superseb:
@PennyScissors @aiyengar2 Can you review to see if this is constructed correctly or if this needs a different approach/layout so I can work on getting it ready to merge

main.go
```diff
@@ -234,6 +240,27 @@ func listPackages(c *cli.Context) {
...
+	}
+	var allImages []string
+	for _, values := range versionValues {
+		chartImages := validate.PickImagesFromValuesMap(values, "dummy")
+		allImages = append(allImages, chartImages...)
+	}
+	if PorcelainMode {
+		fmt.Println(strings.Join(allImages, " "))
+		return
+
```
PennyScissors, line 258: Nit: unnecessary line break

pkg/validate/validate.go
```diff
@@ -189,3 +196,115 @@ func copyAndUnzip(repoFs billy.Filesystem, upstreamPath, localPath string) error
...
+			chartImages := PickImagesFromValuesMap(values, chartNameAndVersion)
+			for _, chartImage := range chartImages {
+				if !imageHasRancherPrefix(chartImage) {
+					return fmt.Errorf("Image [%s] does not start with rancher/\n", chartImage)
+				}
+			}
+		}
+	}
+	return nil
+}
```
PennyScissors, line 225: Are all functions in validate.go except `CheckImageRancherPrefix` and `imageHasRancherPrefix` duplicates from the image pkg in rancher/rancher? Just wondering how much was duplicated.

superseb: Nah they were already modified in the previous version of this PR and I modified them more now with the `list-asset-images` functionality. What might happen is that the way we get images from charts changes, and after changing it in Rancher, we run into it here as well or something but I think that's not a big issue given that the other side is changing Rancher's side and having a dependency there.

superseb:
@aiyengar2 said this will be included in the next charts validation/testing tool

PR #87, opened 2022-10-31
Support ignoreDependencies and replacePaths

aiyengar2:
ignoreDependencies supports the ability to be able to drop dependencies grabbed by upstream in case a chart owner wants to omit such dependencies. For example, for Project Monitoring V2 we would like to omit all exporter charts without storing a ton of patches and exclude files.

replacePaths supports the ability to mark particular paths in a chart as ones that should **not** be stored under `generated-changes/patches` but instead be stored in both `generated-changes/overlay` and `generated-changes/exclude`. While this doesn't change any functionality per-say, it makes the contents of the generated-changes directory more readable.

### DIFF
--- a/pkg/change/apply.go
+++ b/pkg/change/apply.go
@@ -63,22 +63,22 @@ func ApplyChanges(fs billy.Filesystem, toDir, gcRootDir string) error {
 			return err
 		}
 	}
-	exists, err = filesystem.PathExists(fs, chartsOverlayDirpath)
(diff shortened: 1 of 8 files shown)


aiyengar2:
Ran `./scripts/package-ci`, found errors but they aren't due to the changes in this script since they appear in the latest charts-build-scripts too. Noted in https://github.com/rancher/charts-build-scripts/issues/90.

PR #91, opened 2022-12-01
Improve generation of release.yaml file

stormqueen1990:
### Changes
* Change the logic for validation so `release.yaml` does not get wiped out when packages are missing.
* Add semver ordering to `release.yaml`.

### DIFF
--- a/pkg/options/validate.go
+++ b/pkg/options/validate.go
@@ -4,6 +4,9 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/hashicorp/go-version"
+	"golang.org/x/exp/slices"
+
 	"github.com/go-git/go-billy/v5"
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
 	"gopkg.in/yaml.v2"
@@ -41,8 +44,16 @@ func (r ReleaseOptions) Append(chartName string, chartVersion string) ReleaseOpt
 	if !ok {
 		versions = []string{}
 	}
+
+	if slices.Index(versions, chartVersion) != -1 {
+		// value is present, do not include it
+		return r
+	}
+
 	versions = append(versions, chartVersion)
+
 	r[chartName] = versions
+
 	return r
 }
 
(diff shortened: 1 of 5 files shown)


pkg/options/validate.go
```diff
@@ -88,7 +88,7 @@ func (r ReleaseOptions) WriteToFile(fs billy.Filesystem, path string) error {
 	if !exists {
 		file, err = filesystem.CreateFileAndDirs(fs, path)
 	} else {
-		file, err = fs.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, os.ModePerm)
+		file, err = fs.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_APPEND, os.ModePerm)
```
aiyengar2, line 91: Hmm I'm not sure if I agree with this solution.

In the following code block that uses the file:

https://github.com/rancher/charts-build-scripts/blob/a46c0df20db94e46d54c6f4e1b673191df7d93c7/pkg/validate/validate.go#L43-L51

The behavior of the code is to modify the release.yaml to be what the user expects it to be by adding the charts that fall under one of the three categories into the release.yaml:
- `UntrackedInRelease`: charts that are locally listed that are not in the release branch
- `RemovedPostRelease`: charts that are in the release branch but not locally listed
- `ModifiedPostRelease`: charts that are in both the release branch and locally listed but have differing files

If we simply append instead of truncating, wouldn't that result in multiple runs of `make validate` adding the same chart entries multiple times? That doesn't seem right.

aiyengar2 (changes requested):

pkg/options/validate.go
```diff
@@ -74,6 +88,29 @@ func LoadReleaseOptionsFromFile(fs billy.Filesystem, path string) (ReleaseOption
 	return releaseOptions, yaml.Unmarshal(releaseOptionsBytes, &releaseOptions)
 }
 
+// SortBySemver sorts the version strings in release options according to semver constraints
+func (r ReleaseOptions) SortBySemver() {
+	for chartName, versions := range r {
+		slices.SortFunc(versions, CompareVersions)
+		r[chartName] = versions
+	}
+}
```
aiyengar2, line 97: Where does this get used?

pkg/options/validate.go
```diff
@@ -41,8 +44,19 @@ func (r ReleaseOptions) Append(chartName string, chartVersion string) ReleaseOpt
...
+	if slices.Index(versions, chartVersion) != -1 {
+		// value is present, do not include it
+		return r
+	}
+
 	versions = append(versions, chartVersion)
+
+	// keep sorted by semver
+	slices.SortFunc(versions, CompareVersions)
+
```
aiyengar2, line 57: Should we keep it sorted on every append, or should we just call `releaseYaml.SortBySemver()` when we actually dump the release.yaml?

I like the latter since it might be more efficient, unless there's a reason you'd prefer to keep the ReleaseOptions always sorted!

aiyengar2 (commented):
overall, looks good to me. But have questions about the extra function we're leaving behind and doing sorting operations on every append

pkg/options/validate.go
stormqueen1990: Oh I originally had in mind it would be called before saving the file. But then I forgot to actually _make_ the call, will update.

pkg/options/validate.go
stormqueen1990: Yeah that's fair. I have no good reason for keeping sorted, other than expecting the output to be sorted.

pkg/options/validate.go
stormqueen1990: I'm going to resolve this thread since this line has been reverted.

pkg/options/validate.go
stormqueen1990: I removed this and instead added a new call to `SortBySemver()` before dumping the contents into `release.yaml`.

aiyengar2 (approved):
Thanks! LGTM

Issue #18, opened 2021-02-11
Support aliasing with dependency.yaml


aiyengar2:
Currently, rancher-monitoring required us to repackage `packages/pushprox` for every single pushprox dependency (e.g. compare the currently released version in [main](https://github.com/rancher/charts/tree/999fd4788a7cee740732231c3777d0b052638fa3/charts/rancher-monitoring/rancher-monitoring/9.4.202/charts) vs the one that will be released in [dev-v2.5](https://github.com/rancher/charts/tree/0949cebf7c90f705547ba8b02f520f09be88e4c3/charts/rancher-monitoring/rancher-monitoring/9.4.203-rc01/charts)).

Instead, we should be able to specify a dependency.yaml that takes in aliases that are written into the requirements.yaml so that we only need to have rancher-pushprox pulled in once and referred to several times, as we did before.

aiyengar2:
Possibly blocked on, or should be resolved by, https://github.com/rancher/charts-build-scripts/issues/68

Issue #68, opened 2021-11-27
Problem with chart dependencies


eliaskoromilas:
The following line messes with how Helm 3 handles dependencies:

https://github.com/rancher/charts-build-scripts/blob/v0.2.4/pkg/charts/dependencies.go#L257

more at https://github.com/rancher/partner-charts/pull/265

aiyengar2:
Thanks for bringing this up @eliaskoromilas; seems like we need to slightly modify the logic we use to add dependencies to also ensure versions for those dependencies are added.

This will require more investigation, but I think this can be done as part of a new minor release of the charts (since adding the versions to the Chart.yaml will invalidate existing patches recorded in Chart.yaml.patch files, which will require a hand-held migration process).

PR #106, opened 2023-07-30
Handle aliases in chart dependencies

hardys:
It's valid to have multiple dependencies with alternate aliases which currently are removed on chart generation.

One way to resolve this is just to operate directly on the dependencies list instead of generating a map keyed by name, which is less efficient but avoids the problem of removing valid dependencies that use aliases.

Fixes: #18

### DIFF
--- a/pkg/charts/dependencies.go
+++ b/pkg/charts/dependencies.go
@@ -250,35 +250,29 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
 	if err != nil {
 		return err
 	}
-	// Pick up all existing dependencies tracked by Helm by name
-	helmDependencyMap := make(map[string]*helmChart.Dependency, len(dependencyMap))
-	for _, dependency := range chart.Metadata.Dependencies {
-		if _, ok := dependencyMap[dependency.Name]; !ok {
-			continue
-		}
-		helmDependencyMap[dependency.Name] = dependency
-	}
+
 	// Update the Repository for each dependency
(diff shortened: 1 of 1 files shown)


pkg/charts/dependencies.go
```diff
@@ -250,35 +250,28 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
...
-	}
+
 	// Update the Repository for each dependency
 	for dependencyName := range dependencyMap {
-		d, ok := helmDependencyMap[dependencyName]
-		if !ok {
+		found := false
+		for _, d := range chart.Metadata.Dependencies {
+			if d.Name == dependencyName {
+				d.Repository = fmt.Sprintf("file://./charts/%s", dependencyName)
```
rohitsakala, line 259: Did we miss doing d.Version =. ""  ? here wantedly ?

hardys: Good point - I had been looking at https://github.com/rancher/charts-build-scripts/issues/68 and was thinking that this remapping of repositories probably shouldn't mess with any existing version.

But that is an separate issue so I'll reinstate the version-removal logic here, and we can perhaps discuss the version issue separately via a follow-up PR.

rohitsakala: Okay thanks. Please update the PR and make a new PR

hardys: Done!

rohitsakala (approved):

PR #109, opened 2023-08-09
Checking if container images with required tags exist in Docker Hub

lucasmlp:
- Checks if container images with required tags exist in Docker Hub through a HEAD request
- If the request is rate limited, grab the retry after datetime from the header and retries after it
- Maximum number of retries is five

### DIFF
--- a/pkg/images/checkImages.go
+++ b/pkg/images/checkImages.go
@@ -5,10 +5,14 @@ import (
 	"strings"
 
 	"github.com/rancher/charts-build-scripts/pkg/regsync"
+	"github.com/rancher/charts-build-scripts/pkg/rest"
+	"github.com/sirupsen/logrus"
 )
 
 // CheckImages checks if all container images used in charts belong to the rancher namespace
 func CheckImages() error {
+	failedImages := make(map[string][]string, 0)
+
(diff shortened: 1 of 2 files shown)


pkg/images/checkImages.go
```diff
@@ -20,6 +24,27 @@ func CheckImages() error {
 	if len(nonMatchingImages) > 0 {
 		return fmt.Errorf("found images outside the rancher namespace: %v", nonMatchingImages)
 	}
+
+	// Loop through all images and tags to check if they exist
+	for image := range imageTagMap {
+
+		// Spllit image into namespace and repository
```
rohitsakala, line 31: typo

pkg/rest/head.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const maxRetries = 1
```
rohitsakala, line 12: Can we have this as 5 ? Just in case.

lucasmlp: Sure

rohitsakala (approved):

PR #110, opened 2023-08-11
Adding token to Docker Hub requests if credentials are set

lucasmlp:
- Retrieve token from Docker Hub if credential environment variables are set
- Will not fail if credentials are unset or if fails to retrieve a token. Instead, will send unauthenticated requests to check tags
- Execution time is improved from 4'30" to 2'10" if credentials are set

### DIFF
--- a/pkg/rest/post.go
+++ b/pkg/rest/post.go
@@ -1,0 +1,52 @@
+package rest
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// Post sends a POST request to the given URL with the given body and decodes the response into the given response model.
+func Post(url string, body, responseModel any) error {
+
+	// Marshal the body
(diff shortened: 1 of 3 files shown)


pkg/rest/post.go
```diff
@@ -0,0 +1,53 @@
...
+	// Create a new HEAD request
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonBody))
+	if err != nil {
+		return fmt.Errorf("error creating the HEAD request: %v", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+
+	// Create a new HTTP client
+	client := &http.Client{
```
diogoasouza, line 28: You should probably add a timeout here

rohitsakala (changes requested):

rohitsakala (approved):

diogoasouza (approved):

PR #118, opened 2023-10-12
Do not remove dependencies versions

Kristian-ZH:
With this PR the dependencies versions will not be deleted when they are explicitly defined in the Chart.yaml file.
They are sometimes necessary even in the local charts in Helm because, without the versions, the global Values file does not override the subchart values files (that is expected behavior in Helm)

### DIFF
--- a/pkg/charts/dependencies.go
+++ b/pkg/charts/dependencies.go
@@ -253,11 +253,17 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
 
 	// Update the Repository for each dependency
 	for dependencyName := range dependencyMap {
+		componentChart, err := helmLoader.Load(filesystem.GetAbsPath(fs, filepath.Join(mainHelmChartPath, fmt.Sprintf("charts/%s", dependencyName))))
+		if err != nil {
+			return err
+		}
+		componentVersion := componentChart.Metadata.Version
+
 		found := false
 		for _, d := range chart.Metadata.Dependencies {
 			if d.Name == dependencyName {
 				d.Repository = fmt.Sprintf("file://./charts/%s", dependencyName)
-				d.Version = "" // Local chart archives don't need a version
+				d.Version = componentVersion
 				found = true
 			}
 		}
@@ -266,7 +272,7 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
 			d := &helmChart.Dependency{
 				Name:       dependencyName,
 				Condition:  fmt.Sprintf("%s.enabled", dependencyName),
-				Version:    "", // Local chart archives don't need a version
+				Version:    componentVersion,
 				Repository: fmt.Sprintf("file://./charts/%s", dependencyName),
 			}
 			chart.Metadata.Dependencies = append(chart.Metadata.Dependencies, d)


hardys:
I think this is related to https://github.com/rancher/charts-build-scripts/issues/68 - If we make this change it should resolve that issue?

pkg/charts/dependencies.go
```diff
@@ -266,7 +265,7 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
 			d := &helmChart.Dependency{
 				Name:       dependencyName,
 				Condition:  fmt.Sprintf("%s.enabled", dependencyName),
-				Version:    "", // Local chart archives don't need a version
+				Version:    "",
```
hardys, line 268: Is this still needed, or should it be removed in both places?

Kristian-ZH:
Yes, this should resolve the mentioned issue

pkg/charts/dependencies.go
Kristian-ZH: Currently, this is a flow where there is a dependency in the `generated-changes/dependencies` but it is missing in the `Chart.yaml` file. In this case the Version does not exist but I will try to add some additional logic where it will be fetched from the `packages/packageName/Chart.yaml` file

Kristian-ZH: Currently, there is a logic in the `Preperation` step that intentionally skips the local charts and that's why versions are not fetched from there. This seems to be wrong in the concepts of Helm as the issue above points.
I will try to add more detailed logic that handles the local packages properly and will update the PR because this change now is just a workaround

Kristian-ZH: The new update should address the issue

pkg/charts/dependencies.go
```diff
@@ -257,16 +257,21 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
 		for _, d := range chart.Metadata.Dependencies {
 			if d.Name == dependencyName {
 				d.Repository = fmt.Sprintf("file://./charts/%s", dependencyName)
-				d.Version = "" // Local chart archives don't need a version
```
rohitsakala, line 260: d.Version needs a version here as well just like in line 268. I checked d.Version is sometimes empty.

pkg/charts/dependencies.go
```diff
@@ -257,16 +257,21 @@ func UpdateHelmMetadataWithDependencies(fs billy.Filesystem, mainHelmChartPath s
...
 				found = true
 			}
 		}
 		if !found {
+			componentChart, err := helmLoader.Load(filesystem.GetAbsPath(fs, fmt.Sprintf("../%s/charts", dependencyName)))
```
rohitsakala, line 264: This is failing...this should be 

```
componentChart, err := helmLoader.Load(filesystem.GetAbsPath(fs, filepath.Join(mainHelmChartPath, fmt.Sprintf("charts/%s", dependencyName))))
```

rohitsakala (changes requested):

rohitsakala:
TODO(@rohitsakala):

- [ ] Check if this is not a breaking change for all charts in rancher/charts.

rohitsakala (approved):

lucasmlp (commented):
LGTM 🚀

rohitsakala:
fixes https://github.com/rancher/charts-build-scripts/issues/68

PR #124, opened 2024-01-11
Download icons locally when running make charts

diogoasouza:
When running the `make charts` command, it will check if the icon is pointing to a HTTP URL, and if so, download it and change the `Chart.yaml` to generated from the packages folder to point to it

### DIFF
--- a/pkg/helm/export.go
+++ b/pkg/helm/export.go
@@ -3,8 +3,12 @@ package helm
 import (
 	"fmt"
 	"math"
+	"net/url"
 	"os"
 	"path/filepath"
+
+	"github.com/rancher/charts-build-scripts/pkg/icons"
+	"helm.sh/helm/v3/pkg/chartutil"
 
 	"github.com/blang/semver"
 	"github.com/go-git/go-billy/v5"
@@ -26,7 +30,7 @@ var (
 
 // ExportHelmChart creates a Helm chart archive and an unarchived Helm chart at RepositoryAssetDirpath and RepositoryChartDirPath
 // helmChartPath is a relative path (rooted at the package level) that contains the chart.
-func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, packageVersion *int, version *semver.Version, upstreamChartVersion string, omitBuildMetadata bool) error {
+func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, packageVersion *int, version *semver.Version, upstreamChartVersion string, omitBuildMetadata, validateOnly bool) error {
 	// Try to load the chart to see if it can be exported
 	absHelmChartPath := filesystem.GetAbsPath(fs, helmChartPath)
 	chart, err := helmLoader.Load(absHelmChartPath)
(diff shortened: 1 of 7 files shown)


rohitsakala (changes requested):
I have two concerns here 

1. This kind of logic will make the Chart.yaml inside chart tar differ from index.yaml. This voilates one of the checkpoints we have when we release a chart. 
2. We are modifying metadata of something that is already in released and in use for old versions.

Alternate Solution 
- Can we consider a make target which blocks a user in adding a version of a chart that has chart.icon not pointing to file://.assets/logos/{image}.png

pkg/icons/icons.go
```diff
@@ -0,0 +1,80 @@
...
+	absRepositoryHelmIndexFile := filesystem.GetAbsPath(rootFs, path.RepositoryHelmIndexFile)
+	helmIndexFile, err := helmRepo.LoadIndexFile(absRepositoryHelmIndexFile)
+	if err != nil {
+		return fmt.Errorf("encountered error while trying to load existing index file: %s", err)
+	}
+	helmIndexFile.SortEntries()
+	for name := range helmIndexFile.Entries {
+		latestEntryVersion := helmIndexFile.Entries[name][0]
+		latestIconName := name
+		if strings.Contains(name, "-crd") {
```
nicholasSUSE, line 29: // We don't add any icons for "CRD" because they are not displayed as installable charts

pkg/icons/icons.go
```diff
@@ -0,0 +1,80 @@
...
+		if latestIcon == "" || strings.HasPrefix(latestIcon, "file://") {
+			logrus.Infof("Skipping icon download for %s", name)
+			continue
+		}
+		p := fmt.Sprintf("assets/logos/%s", latestIconName)
+		logrus.Infof("Downloading icon download for %s", name)
+		err := download(rootFs, latestIcon, p)
+		if err != nil {
+			logrus.Infof("failed to download icon for %s", name)
+			continue
```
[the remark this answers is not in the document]
nicholasSUSE: @rohitsakala 
We don't want to return an error because sometimes one Icon URL might be broken.
Then we should keep the icon field with the latest successful icon file.

Of course, we will need to do a one-time effort to guarantee we have a single available icon file for all charts. 

From this point on, if the URL is broken we don't need to halt the flow because we still have one icon working.

pkg/icons/icons.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		for _, version := range helmIndexFile.Entries[name] {
+			version.Icon = fmt.Sprintf("file://%s%s", p, ext)
+		}
+	}
+	err = helmIndexFile.WriteFile(absRepositoryHelmIndexFile, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("encountered error while trying to write updated Helm index into index.yaml: %s", err)
```
nicholasSUSE, line 50: I don't know if these errors show which icon string it failed to write, but it should be for our sanity when debugging any issues in this process.

pkg/icons/icons.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return nil
+}
+
+func download(rootFs billy.Filesystem, url, path string) (string, error) {
```
nicholasSUSE, line 55: It is not needed for `rancher/charts`
But for `rancher/partner-charts` if possible and if we are the ones who will implement this same strategy. 
I would consider a for loop with go functions for the download execution. 
120+ rancher partner charts

pkg/icons/icons.go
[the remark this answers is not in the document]
rohitsakala: Okay, gotcha. So we make sure all icons are present first.

rohitsakala (approved):

rohitsakala (approved):

pkg/helm/export.go
```diff
@@ -59,6 +64,27 @@ func ExportHelmChart(rootFs, fs billy.Filesystem, helmChartPath string, packageV
 	}
 	chartVersion := chartVersionSemver.String()
 
+	if releaseOptions != nil && releaseOptions.Contains(chart.Metadata.Name, chartVersion) {
```
rohitsakala, line 67: I tried creating a new version of a chart and was trying to see if the icon URL is changed. 

It is not changing since the new version is not yet in release.yaml. Informing all feature chart owners about adding the version beforehand to release.yaml is too complicated I feel. 

If you don't have any other solution, I am okay with adding a simple make target in CI which will check if the chart.Icon is not set to URL or not and if it is set to http(s) URL, the CI will fail.

rohitsakala (changes requested):

pkg/helm/export.go
diogoasouza: You're right. I changed the code to just use a simple bool flag depending on which command is being ran

rohitsakala (approved):

lucasmlp (approved):
LGTM 🚀

PR #126, opened 2024-01-16
Running make validate only for default branch of rancher/charts repo

lucasmlp:
This pull request changes the previous approach to running `make validate` in all active branches to just the default one.

- Removed the artifact management steps as it is not necessary anymore.
- Merged the `validate` and `build` jobs into `build-validate`
- The `build-validate job` only checks out the default branch. Not specifying the `ref` argument in the `actions/checkout@v3` step causes the step to check out the default branch.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -13,61 +13,24 @@ jobs:
       - name: Run CI
         run: dapper ci
 
-  build-binary:
+  build-validate:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout 
       uses: actions/checkout@v3
-
-    - name: Checkout Pull-Request
-      run: gh pr checkout ${{ github.event.pull_request.number }}
-      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yaml
rohitsakala, line 1: We don't need to gh pr checkout since action/checkout will checkout the changes of the PR. See `build` job for reference.  Please confirm this.

rohitsakala (changes requested):

.github/workflows/pull-request.yaml
lucasmlp: You right.
The actions/checkout step runs `git checkout --progress --force refs/remotes/pull/126/merge` and checks out the state of the repository as if the pull request has been merged, but without actually merging it. It's like a preview of the post-merge state.
The `gh pr checkout 126` checks out the branch associated with pull request #126.

rohitsakala (approved):

PR #127, opened 2024-01-24
adding make icon script instead of doing it inside make charts

diogoasouza:
Extracting the logic of download icons from make charts to a separate command


recena:
@diogoasouza In the future, every PR should be linked / associated to a GH issue, or least, we should have a PR template to provide context, motivation, acceptance criteria, etc...

nicholasSUSE (approved):

PR #128, opened 2024-03-13
Simplify `charts-build-scripts` tooling

### LINKED ISSUE #121: Support ARM builds
For users of new Macs, we only currently publish amd64 binaries.

We should also publish ARM binaries and make the appropriate changes to https://github.com/rancher/charts-build-scripts/blob/3dd99c1209a313fd91b5594f87b044002e6bd3be/templates/template/scripts/pull-scripts#L28-L43 to avoid having such users have to build the binary locally.

Encountered in https://github.com/aiyengar2/Rancher-Plugin-gMSA/pull/8 since the local binary building process doesn't work when you have workspaces.

I have appropriately also filed https://github.com/rancher/charts-build-scripts/pull/120 in order to fix this issue for repositories using workspaces with charts-build-scripts, which should be a good stopgap solution for now.

adamkpickering:
I ran out of things to do, so decided to familiarize myself with rancher/charts-build-scripts. It seems to me that the build process/CI is convoluted. I get the sense that it came from a repository template, and many of the features of that template never were needed by this repository.

Since charts-build-scripts is a plain CLI tool, the tooling doesn't have to be complicated. This PR simplifies it. It uses the go toolchain, `golangci-lint` (which the repository already uses), and `goreleaser` (which is popular and effective - I've used it in the past) for releases.

If others on MAPPS disagree with this way of doing things that is totally fine. I know I didn't put this to the team for discussion, so I may have missed something important. I knew that this wouldn't take very long, and it is sometimes easier to critique a working example rather than explaining how something is going to work.

If we merge this, we will need to modify https://github.com/rancher/charts/blob/dev-v2.9/scripts/pull-scripts.

As a bonus, fixes #121.


adamkpickering:
Please hold off on reviewing this. It uses `goreleaser`, which is not (yet?) on the list of allowed github actions. I will post another message here when I have an update.

adamkpickering:
I confirmed that we can use `goreleaser`. This PR is ready for review.

README.md
```diff
@@ -15,23 +12,53 @@ Export BRANCH_ROLE as `staging`, `live`, or `custom`. Then run:
...
+golangci-lint run ./...
+```
+
+### How to Release
+
+Releases are done via a github action that uses [`goreleaser`] (https://goreleaser.com/).
+In order to release, simply tag the commit that you want the release
+to be based off of with a tag that is in semver format. `goreleaser`
+takes care of the rest.
+
```
nicholasSUSE, line 58: Could you provide a more detailed on the step by step on doing this? 
Imagine someone that never used `gorelease` before and also doesn't know the rules of versioning on this repo.

README.md
```diff
@@ -51,15 +51,15 @@ golangci-lint run ./...
...
+Releases are done via a github action that uses [`goreleaser`](https://goreleaser.com/).
 In order to release, simply tag the commit that you want the release
 to be based off of with a tag that is in semver format. `goreleaser`
 takes care of the rest.
 
 
 ## License
 
-Copyright (c) 2019 [Rancher Labs, Inc.](http://rancher.com)
+Copyright (c) 2024 [Rancher Labs, Inc.](http://rancher.com)
```
nicholasSUSE, line 62: Nice!

.github/workflows/release.yaml
```diff
@@ -4,8 +4,6 @@ on:
   push:
     tags:
       - 'v*'
-    branches:
-      - master
```
nicholasSUSE: Are you sure about this? 
Won't this start running the release workflow for all branches? 
I think we want it only in the master branch, unless you are doing this for testing purposes.

nicholasSUSE (changes requested):

README.md
adamkpickering: Done. Let me know if I can improve any part of it.

.github/workflows/release.yaml
adamkpickering: I think you're right, nice catch. I'll set it so that the workflow only triggers on pushes of branches of the form `v*`.

adamkpickering: This is a confusing part of github actions. I double checked, and the `tags` and `branches` fields work in an OR manner, not AND. Consider the following:
```yaml
  push:
    tags:
      - 'v*'
    branches:
      - master
```
This causes the workflow to be triggered when a tag of the form `v*` is pushed, OR when a commit is pushed to `master`. This isn't what we're looking for.

I agree that it would be better to only trigger this workflow if a tag of the form `v*` was pushed to the `master` branch specifically, but I don't see a way to do this. In practice, it is not likely that we will accidentally push a tag of this form that is not pointing at a commit on `master`.

nicholasSUSE: Ok understood.

nicholasSUSE (approved):

nicholasSUSE (approved):

PR #129, opened 2024-03-17
Support pulling crds from an upstream

### LINKED ISSUE #13: Allow CRD Charts to pull from an upstream
An `AdditionalChart` that has `CRDOptions` should also be able to specify `UpstreamOptions` as long as `templateDirectory` is not set. This should trigger the charts to pull the chart base from a specified upstream and then move CRDs over to a given directory (conflicting filenames will be overridden).

Before releasing, we'll need to confirm generated changes for additional charts works as expected.

Running prepare should error out if a user specifies both `upstream` and `templateDirectory`. Eventually, we can completely deprecate the `templateDirectory` field by just storing the template directly in the `workingDir` of the CRD chart and using `url: local`.

#### Why?

This feature enhancement will allow us to use the `packages/{packageName}` mechanism to point to two new charts: `rancher-crd-chart-base` and `rancher-crd-chart-assume-ownership`.

Once this is done, we'll be able to maintain all of the CRD charts in two central packages instead of copying changes to templates across Backup, CIS, Monitoring, Kiali Server, etc.

joshmeranda:
# Problem

While updating the upstream for `rancher-monitoring` chart, I found that the crd generation code is unable to work with the structure of `kube-prometheus-stack`. The `charts-build-scripts` expect crds to be stored in a `crds` directory at the chart root directory; however, `kube-prometheus-stack` has moved its crds into a [subchart](https://github.com/prometheus-community/helm-charts/tree/7c6906ca223344c06952007fda670c6c81e6d1da/charts/kube-prometheus-stack/charts/crds). This causes several problems for crd generation:

1. Since we clear the `charts` directory when loading dependencies, the upstream crd subchart is removed before the crd is generated, so we cannot simply support a user specified location to look for crds over the default `crds`.
2. The upstream crd subchart is poorly defined so that the version is `0.0.0` and does not match the real upstream version `57.0.3`, so we cannot simply manage a `rancher-monitoring-crd` since the `rancher-monitoring` and `rancher-monitoring-crd` charts will not have matching versions.
3. The current `additionalCharts` configuration validation code does not allow for specifying both `upstreamOptions` and `crdOptions` so we cannot pull from an upstream and then use the crd template.

# Solution

Allow for specifying both `upstreamOptions` to pull the crds directory, and `crdOptions` to use the crd template.

fixes #13 
unblocks [#44614](https://github.com/rancher/rancher/issues/44614)

### DIFF
--- a/pkg/charts/additionalchart.go
+++ b/pkg/charts/additionalchart.go
@@ -117,6 +117,13 @@ func (c *AdditionalChart) Prepare(rootFs, pkgFs billy.Filesystem, mainChartUpstr
 		if err != nil {
 			return fmt.Errorf("encountered error while trying to get the main chart's working directory: %s", err)
 		}
+		if c.Upstream != nil {
+			logrus.Infof("pulling CRD upstream")
+			u := *c.Upstream
+			if err := u.Pull(rootFs, pkgFs, filepath.Join(mainChartWorkingDir, path.ChartCRDDir)); err != nil {
+				return fmt.Errorf("encountered error while trying to pull upstream into %s: %s", mainChartWorkingDir, err)
+			}
+		}
 		exists, err := filesystem.PathExists(pkgFs, filepath.Join(mainChartWorkingDir, path.ChartCRDDir))
 		if err != nil {
 			return fmt.Errorf("encountered error while trying to check if %s exists: %s", filepath.Join(mainChartWorkingDir, path.ChartCRDDir), err)
--- a/pkg/charts/parse.go
+++ b/pkg/charts/parse.go
@@ -162,9 +162,6 @@ func GetChartFromOptions(opt options.ChartOptions) (Chart, error) {
 // GetAdditionalChartFromOptions returns an AdditionalChart based on the options provided
 func GetAdditionalChartFromOptions(opt options.AdditionalChartOptions) (AdditionalChart, error) {
 	var a AdditionalChart
-	if opt.UpstreamOptions != nil && opt.CRDChartOptions != nil {
-		return a, fmt.Errorf("invalid additional chart options provided: cannot define both UpstreamOptions and CRDChartOptions")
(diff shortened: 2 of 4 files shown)


adamkpickering (changes requested):
I think that allowing both `crdOptions` and `upstreamOptions` can work.

But, I don't think these changes are going to work like you want. Take a closer read through `AdditionalChart.Prepare()`: on line 114 there is a check for `c.CRDChartOptions != nil`. If it is not `nil`, then the following `else` clause, which pulls code using the object that results from `upstreamOptions`, does not run. So I think that using this code as you intend will result in an error because there is no `crds/` directory.

Another thing: if you have time, it would be great if you could write some unit tests for your change. `charts-build-scripts` has none at the moment, and that makes it hard to change.

joshmeranda:
@adamkpickering I might be misunderstanding you but you are correct that if `c.CRDChartOptions != nil` we won't enter that next `else` branch to pull the charts. But it will enter the nested `if c.Upstream != nil` check to pull down the crds. This works as expected for `rancher-logging` (no crd upstream) and `rancher-monitoring` (with crd upstream after rebase)

adamkpickering:
Ah you're right, I'm not sure how I missed that. Moving too fast I guess. The code looks good to me then.

We do need to update the documentation and comments though - there are places that say that `crdOptions` is mutually exclusive with `upstreamOptions`. Specifically `templates/template/docs/packages.md` and `pkg/options/additionalchart.go`.

There is duplication of this documentation in rancher/charts due to the templating. Don't worry about this - I will bring this up internal to the MAPPS team, and we'll have to open another PR or make other changes to address this problem.

joshmeranda:
Ope I forgot about the code docs. I'll go through and update what I can in the code. Thanks!

adamkpickering (approved):

PR #133, opened 2024-06-03
Fix make validade lifecycle assets

nicholasSUSE:
### Problem
We have made the `make lifecycle-assets` to enforce the rules for the assets versions deprecation on new branch versions. 

However `make validate` started to complain against these removed assets versions. 

### Solution 
Implement the check for the assets versions that are not in the lifecycle at `make validate`


lucasmlp (approved):
LGTM 🚀

PR #138, opened 2024-06-27
Feature Implementation of Automatic forward port of asset versions, i.e: make auto-forward-port

nicholasSUSE:
### Problem 
Forward porting each asset version individually is an effortful manual work and prone to human error. 

For each asset version, this was the manual process:
1. Check if the asset version is not present at the `production branch` and present at `development branch`. 
2. Open a new local branch 
3. Clean release.yaml 
4. Add and commit. 
5. execute `make forward-port`
6. Add and commit. 

7. Repeat it for every asset version and it's dependencies 
8. Push to a forked repository. 
9. Open a Pull Request to the target `production branch`. 
10. Review and approve PR. 
11. Monitor `git-mirror`. 

### Solution

Implement automation to reduce the amount of manual work and human intervention. 
This solution automates the above mentioned steps: (`1 -> 8`)

### Results

These are the automated `pushes` generated by this automation:

https://github.com/rancher/charts/pull/4137
https://github.com/rancher/charts/pull/4138
https://github.com/rancher/charts/pull/4139
https://github.com/rancher/charts/pull/4140
https://github.com/rancher/charts/pull/4141
https://github.com/rancher/charts/pull/4142
https://github.com/rancher/charts/pull/4143
https://github.com/rancher/charts/pull/4144
https://github.com/rancher/charts/pull/4145
https://github.com/rancher/charts/pull/4146
https://github.com/rancher/charts/pull/4147
https://github.com/rancher/charts/pull/4148
https://github.com/rancher/charts/pull/4149
https://github.com/rancher/charts/pull/4150
https://github.com/rancher/charts/pull/4151
https://github.com/rancher/charts/pull/4152
https://github.com/rancher/charts/pull/4153
https://github.com/rancher/charts/pull/4154
https://github.com/rancher/charts/pull/4155
https://github.com/rancher/charts/pull/4156
https://github.com/rancher/charts/pull/4157
https://github.com/rancher/charts/pull/4158
https://github.com/rancher/charts/pull/4159
https://github.com/rancher/charts/pull/4160


recena:
@nicholasSUSE You can use the DRAFT feature instead of `[wip]` prefix.

pkg/auto/forward.port.go
```diff
@@ -0,0 +1,161 @@
...
+// After each forward-port execution it will add and commit the changes to the git repository.
+// It will push the branch to the remote repository and delete the local branch before moving,
+// to the next Pull Request.
+func (fp *ForwardPort) executeForwardPorts() error {
+	// save the original branch to change back after forward-port
+	originalBranch := fp.git.Branch
+
+	// create log file
+	fpLogs, err := lifecycle.CreateLogs("forward-ported", "")
+	defer fpLogs.File.Close()
```
briandowns, line 100: The error should be checked first.

pkg/auto/forward.port.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	// write log title and header
+	fpLogs.WriteHEAD(fp.VR, "Forward-Ported Assets")
+
+	for asset, pr := range fp.pullRequests {
+		fpLogs.Write(asset, "INFO")
+
+		// open and check if it is clean the git repo
+		err := fp.createNewBranchToForwardPort(pr.branch)
```
briandowns, line 112: It's a good habit to scope lock the `err` value in the `if` clause. Not doing so allows for this value to be resused but can cause problems if there's a situation where the error doesn't necessitate a return later.

pkg/auto/forward.port.go
```diff
@@ -0,0 +1,161 @@
...
+		// clean release.yaml in the new branch
+		if err = prepareReleaseYaml(); err != nil {
+			return err
+		}
+		// git add && commit cleaned release.yaml
+		if err = fp.git.AddAndCommit("cleaning release.yaml"); err != nil {
+			return err
+		}
+
+		fpLogs.Write(fmt.Sprintf("Branch: %s", pr.branch), "INFO")
```
briandowns, line 126:
```suggestion
		fpLogs.Write("Branch: " + pr.branch, "INFO")
```

When concatenating strings, there's no need to use any of the XprintfX functions. They're incredibly heavy due to the amount of reflection involved.

pkg/git/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	if urlPart1 == "https://github.com" && urlPart2 == repo {
+		return true
+	}
+
+	return false
+}
+
+// extractCommonParts takes two Git URLs and returns the common prefix and suffix.
+func extractCommonParts(url1, url2 string) (string, string) {
+	// Split the URLs into segments
```
briandowns, line 19: Is there ever a chance that a string could be passed in that is empty or doesn't contain a `/`?

pkg/lifecycle/lifecycle.go
```diff
@@ -63,22 +65,30 @@ func InitDependencies(rootFs billy.Filesystem, branchVersion string, currentChar
 		DisableQuote: true,
 	})
 	var err error
+
+	workDir, _ := os.Getwd()
```
briandowns, line 69: What happens if this errors?

pkg/lifecycle/lifecycle.go
```diff
@@ -63,22 +65,30 @@ func InitDependencies(rootFs billy.Filesystem, branchVersion string, currentChar
...
-	clean, err := dep.checkIfGitIsCleanWrapper(debug)
-	if !clean {
-		return nil, fmt.Errorf("git is not clean, it must be clean before proceeding with removing charts")
-	}
+	clean, err := dep.statusPorceLainWrapper(debug)
 	if err != nil {
 		return nil, err
 	}
+	if !clean {
+		return nil, fmt.Errorf("git repo should be clean")
```
briandowns, line 90:
```suggestion
		return nil, errors.New("git repo should be clean")
```

pkg/git/utils.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// CheckForValidForkRemote checks if the remote URL is a valid remote fork for the upstream URL.
+func CheckForValidForkRemote(upstreamURL, remoteURL, repo string) bool {
+	urlPart1, urlPart2 := extractCommonParts(upstreamURL, remoteURL)
+	if urlPart1 == "https://github.com" && urlPart2 == repo {
+		return true
+	}
+
+	return false
```
briandowns, line 14: We can simplify the entire body of this funtion:

```go
return urlPart1 == "https://github.com" && urlPart2 == repo
```

pkg/git/utils.go
```diff
@@ -0,0 +1,48 @@
...
+// extractCommonParts takes two Git URLs and returns the common prefix and suffix.
+func extractCommonParts(url1, url2 string) (string, string) {
+	// Split the URLs into segments
+	segments1 := strings.Split(url1, "/")
+	segments2 := strings.Split(url2, "/")
+
+	// Find the common prefix
+	prefixLength := 0
+	for i := 0; i < len(segments1) && i < len(segments2); i++ {
+		if segments1[i] == segments2[i] {
```
briandowns, line 26: This and the ones below can be simplified:

```go
if segments1[i] != segments2[i] {
    break
}
prefixLength++
```

pkg/git/models.go
```diff
@@ -0,0 +1,8 @@
+package git
+
+// Git struct holds necessary data to work with the current git repository
+type Git struct {
```
briandowns, line 4: This is somewhat of an anti pattern in go. Remove this type and put where it's used. Ie. top of the git.go file.

pkg/lifecycle/logs.go
```diff
@@ -9,17 +9,22 @@ import (
...
+// CreateLogs creates a new log file and returns a logs struct with the file and file path
+func CreateLogs(fileName, detail string) (*Logs, error) {
+	// get a timestamp
+	currentTime := time.Now()
+	now := currentTime.Format("2006-01-02T15:04")
+
+	filePath := fmt.Sprintf("logs/%s_%s_%s", now, detail, fileName)
 
 	// Create the logs directory if it doesn't exist
 	err := os.MkdirAll("logs", 0755)
```
briandowns, line 30: Where is this error checked?

pkg/lifecycle/logs.go
```diff
@@ -33,79 +38,79 @@ func createLogs(fileName string) (*logs, error) {
...
+	l.Write(currentTime.Format("2006-01-02 15:04:05"), "INFO")
+	l.Write(fmt.Sprintf("Branch Version: %.1f", versionRules.BranchVersion), "INFO")
+	l.Write(fmt.Sprintf("minimal version: %d", versionRules.MinVersion), "INFO")
+	l.Write(fmt.Sprintf("max version: %d", versionRules.MaxVersion), "INFO")
+	l.Write(fmt.Sprintf("development branch: %s", versionRules.DevBranch), "INFO")
+	l.Write(fmt.Sprintf("production branch: %s", versionRules.ProdBranch), "INFO")
 
-	rules := make(map[string]string, len(versionRules.rules))
-	for k, v := range versionRules.rules {
+	rules := make(map[string]string)
```
briandowns, line 56:
```suggestion
	rules := make(map[string]string, len(versionRules.Rules))
```

pkg/lifecycle/versions.rules.go
```diff
@@ -146,5 +146,5 @@ func (vr *VersionRules) CheckChartVersionToRelease(chartVersion string) (bool, e
 		logrus.Errorf("failed to check version to release for chartVersion:%s with error:%v", chartVersion, err)
 		return false, err
 	}
-	return chartVersionInt == (vr.maxVersion - 1), nil
+	return chartVersionInt == (vr.MaxVersion - 1), nil
```
briandowns, line 149: The parenthesis aren't necessary.

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+	yqDirPath := filepath.Dir(yqPath)
+	// Append the yq directory to the PATH
+	currentPath := os.Getenv("PATH")
+	newPath := fmt.Sprintf("%s:%s", yqDirPath, currentPath)
+	return newPath, nil
+}
+
+// createForwardPortCommands will create the forward port script commands for each asset and version,
+// and return a sorted slice of commands
+func (fp *ForwardPort) createForwardPortCommands(chart string) ([]Command, error) {
```
briandowns, line 43: Prefer using a single character for method receivers.

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+	* make forward-port
+	* CHART=rancher-provisioning-capi
+	* VERSION=100.0.0+up0.0.1
+	* BRANCH=dev-v2.9
+	* UPSTREAM=upstream
+	 */
+
+	upstreamRemote, ok := fp.git.Remotes["https://github.com/rancher/charts"]
+	if !ok {
+		logrus.Error("upstream remote not found; you need to have the upstream remote configured in your git repository (https://github.com/rancher/charts)")
```
briandowns, line 92: This should be made into a value that can be printed here and returned below rather than being repeated.

```go
var errUpstreamNotFound = errors.New("upstream remote not found; you need to have the upstream remote configured in your git repository (https://github.com/rancher/charts)")
```

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+	return Command{Chart: asset, Command: commands, Version: version}, nil
+}
+
+// checkIfChartChanged will check if the chart has changed from the last chart.
+// It will return true if the chart has changed, otherwise it will return false.
+// If the chart has changed, it will check for special cases like (fleet and neuvector), and CRD dependencies.
+func checkIfChartChanged(lastChart, currentChart string) bool {
+	// Check if the current chart is different from the last chart
+	sameCharts := (lastChart == currentChart)
+	if sameCharts {
```
briandowns, line 114: This can be simplified: Remove line 113 and add the conditional check to line 114.

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+		return false
+	}
+
+	// Check for special edge cases
+	edgeCase := checkEdgeCasesIfChartChanged(lastChart, currentChart)
```
briandowns, line 119: I think this can also be simplified as:

```go
return !checkEdgeCasesIfChartChanged(lastChart, currentChart)
```

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+		if lastParts[i] == currentParts[i] {
+			equalCounter++
+		}
+	}
+
+	if equalCounter == 0 {
+		return true
+	}
+	if equalCounter == 1 && strings.HasPrefix("rancher-", lastParts[0]) &&
+		minLength < 3 {
```
briandowns, line 159: Move tyhis up a line. Go doesn't care at all about line length, per Rob Pike.

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+
+// createNewBranchToForwardPort will create a new branch to forward-port the assets
+func (fp *ForwardPort) createNewBranchToForwardPort(branch string) error {
+	// check if git is clean and branch is up-to-date
+	err := fp.git.IsClean()
+	if err != nil {
+		return err
+	}
+	// create new branch and checkout
+	err = fp.git.CreateAndCheckoutBranch(branch)
```
briandowns, line 190:
```suggestion
	return fp.git.CreateAndCheckoutBranch(branch)
```

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+	_, err := os.Stat("release.yaml")
+	if os.IsNotExist(err) {
+		logrus.Error("release.yaml does not exist.")
+		return fmt.Errorf("release.yaml does not exist")
+	} else if err != nil {
+		return err // return any other error encountered
+	}
+
+	// File exists, open it with O_TRUNC to erase its content
+	file, err := os.OpenFile("release.yaml", os.O_WRONLY|os.O_TRUNC, 0644)
```
briandowns, line 211:
```suggestion
if err := os.Truncate("release.yaml, 0); err != nil {
    return err
}
```

The defer can be removed.

pkg/auto/auto.go
```diff
@@ -0,0 +1,237 @@
...
+		return err
+	}
+	defer file.Close()
+
+	logrus.Info("Content of release.yaml erased successfully.")
+	return nil
+}
+
+// executeCommand will execute the given command using the yqPath if needed
+func executeCommand(command []string, yqPath string) error {
```
briandowns, line 222: Why aren't we using any of the yaml parser libraries available?

pkg/auto/forward.port.go
```diff
@@ -0,0 +1,161 @@
+package auto
```
briandowns, line 1: This file isn't named according to Go standards. Replace the first `.` with an `_`. And do this for any other file names similarly.

main.go
```diff
@@ -43,6 +44,8 @@ const (
 	DefaultDebugEnvironmentVariable = "DEBUG"
 	// DefaultBranchVersionEnvironmentVariable is the default environment variable that indicates the branch version to compare against
 	DefaultBranchVersionEnvironmentVariable = "BRANCH_VERSION"
+	// DefaultForkEnvironmentVariable is the default environment variable that indicates the fork URL
```
briandowns, line 47: None of these need to be exported since they can't be imported from main.

main.go
```diff
@@ -590,36 +619,72 @@ func checkRCTagsAndVersions(c *cli.Context) {
...
+	_, err = lifeCycleDep.CheckLifecycleStatusAndSave(CurrentChart)
 	if err != nil {
 		logrus.Fatalf("Failed to check lifecycle status: %s", err)
 	}
 }
+
+func autoForwardPort(c *cli.Context) {
+
+	if ForkURL == "" {
+		logrus.Fatalf("FORK environment variable must be set to run auto-forward-port")
```
briandowns, line 659:
```suggestion
		logrus.Fatal("FORK environment variable must be set to run auto-forward-port")
```

main.go
```diff
@@ -590,36 +619,72 @@ func checkRCTagsAndVersions(c *cli.Context) {
...
+	}
+
+	// Initialize dependencies with branch-version and current chart
+	logrus.Info("Initializing dependencies for auto-forward-port")
+	rootFs := filesystem.GetFilesystem(getRepoRoot())
+	lifeCycleDep, err := lifecycle.InitDependencies(rootFs, c.String("branch-version"), CurrentChart, false)
+	if err != nil {
+		logrus.Fatalf("encountered error while initializing dependencies: %v", err)
```
briandowns, line 667: Should we also be returning here too?

pkg/auto/forward.port.go
nicholasSUSE: Fixing.

pkg/auto/forward.port.go
nicholasSUSE: Fixing

pkg/auto/forward.port.go
nicholasSUSE: Fixing

pkg/git/utils.go
nicholasSUSE: @briandowns 
Actually, no, in the main.go file; we have this:

```
func autoForwardPort(c *cli.Context) {

	if ForkURL == "" {
		logrus.Fatalf("FORK environment variable must be set to run auto-forward-port")
	}
	
```
But I am adding an extra check before extractCommonParts.

pkg/lifecycle/lifecycle.go
nicholasSUSE: Fixing

pkg/git/utils.go
nicholasSUSE: Keen eyes, fixing.

pkg/git/utils.go
nicholasSUSE: Fixing

pkg/git/models.go
nicholasSUSE: Fixing also will fix the auto package as well.

pkg/lifecycle/logs.go
nicholasSUSE: Fixing

pkg/lifecycle/versions.rules.go
nicholasSUSE: I agree it is not necessary. 

I intentionally wrote this way for easier reading. 
Can we keep this one?

pkg/auto/auto.go
nicholasSUSE: Fixing also for forward.port.go

pkg/auto/auto.go
nicholasSUSE: Fixing

pkg/auto/auto.go
nicholasSUSE: Fixing

pkg/auto/auto.go
nicholasSUSE: Fixing

pkg/auto/auto.go
nicholasSUSE: Fixing

pkg/auto/auto.go
nicholasSUSE: Fixing

pkg/auto/auto.go
nicholasSUSE: Wooow, did not know that. 
Fixing.

pkg/auto/auto.go
nicholasSUSE: Agreed, I will do some research on this.

pkg/auto/forward.port.go
nicholasSUSE: Fixing.

main.go
nicholasSUSE: Fixing

main.go
nicholasSUSE: The entire project was made following this standard of `fatal erroring` when an error reaches the function called by the CLI. 

I need to find out which architecture this CLI app is following. I did not spend time on it; I just followed through with what was in place. 

That being said, this function does not allow for returning errors. 
I need to spend time understanding this. 
But I would prefer to do that after the 2.9.0 release cycle.

pkg/auto/auto.go
nicholasSUSE: @briandowns I had to dig a little bit here. 

The `make forward-port` command calls a bash script in the [charts repo/scripts](https://github.com/rancher/charts/tree/dev-v2.9/scripts). It needs `yq` to operate. 

This feature has automated the `make forward-port` command. 

Can we create a separate issue for migrating the bash script `forward-script` inside the `charts-build-scripts`? 
Otherwise the scope of this feature will increase too much. 
And then I will remove the `os.Exec` commands for `yq` and do everything through go code.

main.go
```diff
@@ -71,6 +74,8 @@ var (
 	CacheMode = false
```
briandowns, line 74: In another pr this should be updated to match the ones above using the 0 value declaration rather than this declaration to false. The 0 value for bools is false but moreover like code should be consistent. Which means the new variable below should be defined as type string since the 0 value for string is "".

briandowns: Will also need to be unexported.

main.go
briandowns: Ah, I see it now.

pkg/auto/auto.go
```diff
@@ -0,0 +1,264 @@
...
+	// create new branch and checkout
+	return f.git.CreateAndCheckoutBranch(branch)
+}
+
+// prepareReleaseYaml will prepare the release.yaml file by erasing its content,
+// this is a good practice before releasing or forward-porting any charts with it.
+func prepareReleaseYaml() error {
+	// Check if the file exists
+	_, err := os.Stat("release.yaml")
+	if os.IsNotExist(err) {
```
briandowns, line 226: The nil check on err should always be done first and if it's not nil, then inside that scope further error checking/handling should occur.

pkg/auto/auto.go
briandowns: Let's do this work in another PR at some point.

pkg/git/utils.go
briandowns: Not necessary if there's no chance.

pkg/git/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	"strings"
+)
+
+// CheckForValidForkRemote checks if the remote URL is a valid remote fork for the upstream URL.
+func CheckForValidForkRemote(upstreamURL, remoteURL, repo string) bool {
+	if remoteURL == "" {
+		return false // Remote URL is empty
+	}
+	urlPart1, urlPart2 := extractCommonParts(upstreamURL, remoteURL)
+	return (urlPart1 == "https://github.com" && urlPart2 == repo)
```
briandowns, line 13:
```suggestion
	return urlPart1 == "https://github.com" && urlPart2 == repo
```

main.go
nicholasSUSE: Fixing

main.go
nicholasSUSE: I created this issue so we don't lose track of this technical debt.
See if you agree please.

https://github.com/rancher/ecm-distro-tools/issues/438

pkg/auto/auto.go
nicholasSUSE: Fixing

pkg/git/utils.go
nicholasSUSE: Fixing

briandowns (approved):

Issue #142, opened 2024-07-26
Patch creates unwanted overlay when CRDs have an upstream


joshmeranda:
Introduced in https://github.com/rancher/charts-build-scripts/pull/129 when running `make patch` on a chart with a CRD chart that pulls from an upstream source, an unexpected `generated-changes/overlay/crds` is created. While this does not affect `make prepare` because we delete the CRDs from the main chart, this does clutter the `generated-changes` directory with unnecessary files and requires developers manually removing the directory after each `make patch` to avoid this.

## To Reproduce

```
> git switch dev-v2.9
Switched to branch 'dev-v2.9'
Your branch is up to date with 'origin/dev-v2.9'.
> USE_CACHE=1 PACKAGE=rancher-monitoring/rancher-monitoring make prepare
...
> USE_CACHE=1 PACKAGE=rancher-monitoring/rancher-monitoring make patch
...
> git status
On branch dev-v2.9
Your branch is up to date with 'origin/dev-v2.9'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        packages/rancher-monitoring/rancher-monitoring/charts-crd/
        packages/rancher-monitoring/rancher-monitoring/charts/
        packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/crds/

nothing added to commit but untracked files present (use "git add" to track)

nicholasSUSE:
Resolved by this: https://github.com/rancher/charts-build-scripts/issues/130

PR #143, opened 2024-07-26
Fix crd patches

joshmeranda:
## Problem

When generating patches for a chart with crds we currently copy the crds pulled from upstream back into the main chart which are then included in the main chart patch.

## Solution

We can't simply check for `c.Upstream == nil` before copying since we will not be able to generate `validate-install-crd.yaml` when `c.CRDOptions.UseTarArchive` is `true` since there won't be any crds to generate the validation on. To fix this, we always generate the validation file based on the crds in the generated chart and only copy those crds to the main chart if there is no upstream.

Unfortunately we have to unarchive the CRDs whenever we need to generate the validation file, but in my testing this does not affect the runtime much. I think the better separation between the `Prepare`, `ApplyMainChanges`, and `RevertMainChanges` makes up for it in terms of readbility. (of course open to disagreement on this point)

Issue: https://github.com/rancher/charts-build-scripts/issues/142

### DIFF
--- a/pkg/charts/additionalchart.go
+++ b/pkg/charts/additionalchart.go
@@ -48,8 +48,10 @@ func (c *AdditionalChart) ApplyMainChanges(pkgFs billy.Filesystem) error {
 	if err != nil {
 		return fmt.Errorf("encountered error while trying to get the main chart's working directory: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


joshmeranda:
@rancher/release-team can I get a review on this?

nicholasSUSE:
Original issue was closed

PR #145, opened 2024-07-30
Create release command

nicholasSUSE:
This Pull Request is for replacing: [forward-port script](https://github.com/rancher/charts/blob/dev-v2.9/scripts/forward-port)

With Golang safe code that can be used within [ecm-distro-tools](https://github.com/rancher/ecm-distro-tools)

Issue: https://github.com/rancher/ecm-distro-tools/issues/446

### DIFF
--- a/pkg/auto/release.go
+++ b/pkg/auto/release.go
@@ -1,7 +1,6 @@
 package auto
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -169,22 +168,4 @@ func removeDuplicates(slice []string) []string {
 	}
 
 	return result
-}
-
-// enforceYamlStandard adds indentation to list items in a YAML string and a new line at the end of the file.
-func enforceYamlStandard(yamlBytes []byte) []byte {
-	var indentedBuffer bytes.Buffer
-	lines := bytes.Split(yamlBytes, []byte("\n"))
-	for i, line := range lines {
-		// Check if the line is not the last empty line after splitting
-		if i != len(lines)-1 {
-			if bytes.HasPrefix(line, []byte("- ")) {
-				indentedBuffer.Write([]byte("  ")) // Add two spaces of indentation
-			}
-			indentedBuffer.Write(line)
-			indentedBuffer.WriteByte('\n')
-		}
-	}
-	// Ensure only one newline at the end
-	return append(indentedBuffer.Bytes(), '\n')
 }
--- a/pkg/auto/release.go
+++ b/pkg/auto/release.go
@@ -4,13 +4,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
 	"github.com/rancher/charts-build-scripts/pkg/git"
 	"github.com/rancher/charts-build-scripts/pkg/lifecycle"
 	"github.com/rancher/charts-build-scripts/pkg/path"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 )
 
 // Release holds necessary metadata to release a chart version
@@ -24,9 +25,6 @@ type Release struct {
 	ReleaseYamlPath string
 	ForkRemoteURL   string
 }
-
-// ReleaseVersions represents the versions within the release.yaml file
-type ReleaseVersions map[string][]string
 
 // InitRelease will create the Release struct with access to the necessary dependencies.
 func InitRelease(d *lifecycle.Dependencies, s *lifecycle.Status, v, c, f string) (*Release, error) {
@@ -63,8 +61,8 @@ func InitRelease(d *lifecycle.Dependencies, s *lifecycle.Status, v, c, f string)
 	r.AssetPath, r.AssetTgz = mountAssetVersionPath(r.Chart, assetVersion)
 
 	// Check again if the asset was already released in the local repository
-	if exist := checkAssetReleased(r.AssetPath); exist {
-		return nil, errors.New("asset already released for chart:" + r.Chart + " version:" + r.ChartVersion)
+	if err := checkAssetReleased(r.AssetPath); err != nil {
+		return nil, fmt.Errorf("failed to check for chart:%s ; err: %w", r.Chart, err)
 	}
 
 	// Check if we have a release.yaml file in the expected path
--- a/pkg/auto/release.go
+++ b/pkg/auto/release.go
@@ -1,0 +1,181 @@
+package auto
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/rancher/charts-build-scripts/pkg/filesystem"
+	"github.com/rancher/charts-build-scripts/pkg/git"
+	"github.com/rancher/charts-build-scripts/pkg/lifecycle"
+	"github.com/rancher/charts-build-scripts/pkg/path"
+	"gopkg.in/yaml.v2"
+)
+
+// Release holds necessary metadata to release a chart version
+type Release struct {
+	git             *git.Git
+	VR              *lifecycle.VersionRules
+	AssetTgz        string
+	AssetPath       string
+	ChartVersion    string
+	Chart           string
+	ReleaseYamlPath string
+	ForkRemoteURL   string
+}
+
+// ReleaseVersions represents the versions within the release.yaml file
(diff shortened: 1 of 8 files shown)


pkg/git/git.go
```diff
@@ -263,3 +278,26 @@ func (g *Git) DeleteBranch(branch string) error {
...
+func (g *Git) CheckFileExists(file, branch string) error {
+	upstreamRemote := g.Remotes["https://github.com/rancher/charts"]
+	target := upstreamRemote + "/" + branch + ":" + file
+
+	// Corrected command to only include the necessary arguments
+	cmd := exec.Command("git", "-C", g.Dir, "cat-file", "-e", target)
+	if err := cmd.Run(); err != nil {
+		return err // Return the error if the file does not exist or any other git error occurs
+	}
+	return nil // Return nil if the file exists
```
rafaelbreno, line 292: I think we can instead directly return the `cmd.Run()` call.
```suggestion
	return exec.Command("git", "-C", g.Dir, "cat-file", "-e", target).Run()
```

pkg/git/git.go
```diff
@@ -263,3 +278,26 @@ func (g *Git) DeleteBranch(branch string) error {
...
+}
+
+// ResetHEAD resets the HEAD of the git repository
+// ex: git reset HEAD
+func (g *Git) ResetHEAD() error {
+	cmd := exec.Command("git", "-C", g.Dir, "reset", "HEAD")
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+	return nil
```
rafaelbreno, line 302: Same here.
```suggestion
	return exec.Command("git", "-C", g.Dir, "reset", "HEAD").Run()
```

rafaelbreno (approved):
Just a couple of comments, besides that LGTM!

Approved as they are non-blockers.

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+	AssetTgz        string
+	AssetPath       string
+	ChartVersion    string
+	Chart           string
+	ReleaseYamlPath string
+	ForkRemoteURL   string
+}
+
+// ReleaseVersions represents the versions within the release.yaml file
+type ReleaseVersions map[string][]string
```
briandowns, line 29: I don' think this needs to be a named type and it doesn't need to be exported.

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/rancher/charts-build-scripts/pkg/filesystem"
+	"github.com/rancher/charts-build-scripts/pkg/git"
+	"github.com/rancher/charts-build-scripts/pkg/lifecycle"
+	"github.com/rancher/charts-build-scripts/pkg/path"
+	"gopkg.in/yaml.v2"
+)
+
+// Release holds necessary metadata to release a chart version
+type Release struct {
```
briandowns, line 17: Is this used outside of this package? If not, please unexport it.

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if err := r.git.CheckFileExists(r.AssetPath, r.VR.DevBranch); err != nil {
+		return fmt.Errorf("asset version not found in dev branch: %w", err)
+	}
+
+	if err := r.git.CheckoutFile(r.VR.DevBranch, r.AssetPath); err != nil {
+		return err
+	}
+
+	if err := r.git.ResetHEAD(); err != nil {
```
briandowns, line 94:
```suggestion
	return r.git.ResetHEAD(); err != nil {
```

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func checkAssetReleased(chartVersion string) bool {
```
briandowns, line 101: Let's change this from bool to error. We shouldn't let errors get ignored.

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return true
+}
+
+func mountAssetVersionPath(chart, version string) (assetPath string, assetTgz string) {
+	// example: assets/longhorn/longhorn-100.0.0+up0.0.0.tgz
+	assetTgz = chart + "-" + version + ".tgz"
+	assetPath = "assets/" + chart + "/" + assetTgz
+	return
```
briandowns, line 113:
```suggestion
```

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// UpdateReleaseYaml reads and parse the release.yaml file to a struct, appends the new version and writes it back to the file.
+func (r *Release) UpdateReleaseYaml() error {
+	var releaseVersions ReleaseVersions
+
+	rvBytes, err := os.ReadFile(r.ReleaseYamlPath)
```
briandowns, line 120:
```suggestion
	b, err := os.ReadFile(r.ReleaseYamlPath)
```

pkg/auto/release.go
briandowns, line 94:
```suggestion
	return r.git.ResetHEAD(); err != nil {
```

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+	updatedYAML, err := yaml.Marshal(releaseVersions)
+	if err != nil {
+		return err
+	}
+
+	// Post-process YAML to adjust indentation for list items
+	updatedYAMLIndented := enforceYamlStandard(updatedYAML)
+
+	// Write the updated YAML back to the file
+	if err := os.WriteFile(r.ReleaseYamlPath, updatedYAMLIndented, 0644); err != nil {
```
briandowns, line 143:
```suggestion
	return os.WriteFile(r.ReleaseYamlPath, updatedYAMLIndented, 0644); err != nil {
```

pkg/auto/release.go
briandowns, line 120: Rather than reading directly into memory, we should use the stream interface methods of encoder and decoder and we can condense the operations down as well as read the stream of bytes. This goes for the unmarshal function below.

pkg/auto/release.go
```diff
@@ -0,0 +1,181 @@
...
+			seen[val] = struct{}{}       // mark string as seen
+			result = append(result, val) // append to result if not seen before
+		}
+	}
+
+	return result
+}
+
+// enforceYamlStandard adds indentation to list items in a YAML string and a new line at the end of the file.
+func enforceYamlStandard(yamlBytes []byte) []byte {
```
briandowns, line 166: Is there no library for formatting or validating?

pkg/git/git.go
```diff
@@ -170,6 +170,20 @@ func (g *Git) CheckoutBranch(branch string) error {
...
 }
 
+// CheckoutFile checks out a file in a branch
+// ex: git checkout <remote>/<branch> -- <file>
+func (g *Git) CheckoutFile(branch, file string) error {
+	upstreamRemote := g.Remotes["https://github.com/rancher/charts"]
+	targetBranch := upstreamRemote + "/" + branch
+
+	cmd := exec.Command("git", "-C", g.Dir, "checkout", targetBranch, "--", file)
+	if err := cmd.Run(); err != nil {
```
briandowns, line 180:
```suggestion
	return cmd.Run(); err != nil {
```

pkg/lifecycle/lifecycle.go
```diff
@@ -22,7 +22,7 @@ type Asset struct {
 // Dependencies holds the necessary filesystem,
 // assets versions map, version rules and methods to apply the lifecycle rules in the target branch
 type Dependencies struct {
-	rootFs            billy.Filesystem
+	RootFs            billy.Filesystem
```
briandowns, line 25: Is this field accessed outside of this package?

pkg/auto/release.go
nicholasSUSE: It is not yet used but will. 

Either way, I am converting it to a private struct.

pkg/auto/release.go
nicholasSUSE: This was removed. 
I am using `"gopkg.in/yaml.v3"` now.

pkg/git/git.go
nicholasSUSE: Refactored all functions here

pkg/lifecycle/lifecycle.go
nicholasSUSE: Yes it is used by auto package

PR #148, opened 2024-09-03
Refactoring lifecycle package and dependencies 

nicholasSUSE:
- Refactoring the code
- Removing debug feature 
- Removing `enforce-lifecycle` command, since it was used only once with several limitations due to not all charts following the chart versioning standards. 
- Improved versioning calculations for version rules to adapt to the new charts versions (i.e: 2.7; 2.8; 2.9; 2.`10`...)
- Updating packages
- Adding new unit-tests


pkg/lifecycle/lifecycle.go
```diff
@@ -2,15 +2,17 @@ package lifecycle
...
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
 	"github.com/rancher/charts-build-scripts/pkg/git"
 	"github.com/rancher/charts-build-scripts/pkg/path"
-	"github.com/sirupsen/logrus"
+)
+
+var (
+	errorChartRepository error = errors.New("chart repository is in an inconsistent state")
```
briandowns, line 14: You don't have to indicate type here. The compiler will infer it. The name should be `errChartRepository`.

pkg/lifecycle/lifecycle.go
```diff
@@ -23,48 +25,19 @@ type Asset struct {
...
-	assetsVersionsMap map[string][]Asset
+	AssetsVersionsMap map[string][]Asset
 	VR                *VersionRules
 	Git               *git.Git
-	// These wrappers are used to mock the filesystem and git status in the tests
-	walkDirWrapper         WalkDirFunc
-	makeRemoveWrapper      MakeRemoveFunc
-	statusPorceLainWrapper StatusPorcelainFunc
-	addAndCommitWrapper    AddAndCommitFunc
+	walkDirWrapper    WalkDirFunc // Used to mock the filesystem in unit-tests
```
briandowns, line 31: This is fine since it's in this repo but generally code for testing should never exist in the implementation code.

pkg/lifecycle/lifecycle.go
```diff
@@ -79,171 +52,58 @@ func InitDependencies(rootFs billy.Filesystem, branchVersion string, currentChar
...
 		return err
 	}
-	// Sort the versions before removing
-	ld.sortAssetsVersions()
-
-	// Execute make remove for each chart and version that is not in the lifecycle.
-	// Commit after each chart removal.
-	removedAssetsVersions, err := ld.removeAssetsVersions(debug)
+	if !exists {
+		return errorChartRepository
```
briandowns, line 99: Is this error used only here? If so, you don't need to predeclare it above.

pkg/lifecycle/parser_test.go
```diff
@@ -159,15 +159,15 @@ func Test_sortAssetsVersions(t *testing.T) {
 	dep.sortAssetsVersions()
 
 	// Assert
-	if dep.assetsVersionsMap["key1"][0].Version != "0.0.1" ||
-		dep.assetsVersionsMap["key1"][1].Version != "0.1.0" ||
-		dep.assetsVersionsMap["key1"][2].Version != "1.0.0" {
+	if dep.AssetsVersionsMap["key1"][0].Version != "0.0.1" ||
```
briandowns, line 162: We don't need to change this now nor do I know if you even wrote this but for reference, Hungarian notation in Go is a code smell and should be avoided.

pkg/lifecycle/version_rules.go
```diff
@@ -0,0 +1,195 @@
...
+	MaxVersion       int                `json:"max-version,omitempty"`
+	DevBranch        string             `json:"dev-branch,omitempty"`
+	DevBranchPrefix  string             `json:"dev-branch-prefix"`
+	ProdBranch       string             `json:"prod-branch,omitempty"`
+	ProdBranchPrefix string             `json:"prod-branch-prefix"`
+}
+
+// loadFromJSON will load the version rules from version_rules.json file at charts repository
+func loadFromJSON(fs billy.Filesystem) (*VersionRules, error) {
+	vr := &VersionRules{
```
briandowns, line 42: This should be moved down to where it's used.

pkg/lifecycle/version_rules.go
```diff
@@ -0,0 +1,195 @@
...
+//	Branch can only hold until 2 previous versions of the current branch version.
+//	Branch cannot hold versions from newer branches, only older ones.
+//
+// See CheckChartVersionForLifecycle() for more details.
+func (v *VersionRules) getMinMaxVersionInts() error {
+	// e.g: 2.9 - 0.2 = 2.7
+	minVersionStr := v.Rules[(branchVersionMinorSum(v.BranchVersion, -2))].Min
+	maxVersionStr := v.Rules[v.BranchVersion].Max
+
+	var err error
```
briandowns, line 105: Error doesn't need and shouldn't be predeclared.

pkg/lifecycle/version_rules.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		logrus.Errorf("failed to check version to release for chartVersion:%s with error:%v", chartVersion, err)
+		return false, err
+	}
+	return chartVersionInt == (v.MaxVersion - 1), nil
+}
+
+// CheckForRCVersion checks if the chart version contains the "-rc" string indicating a release candidate version.
+func (v *VersionRules) CheckForRCVersion(chartVersion string) bool {
+	if strings.Contains(strings.ToLower(chartVersion), "-rc") {
```
briandowns, line 173: This can be simplified to:

```go
return strings.Contains(strings.ToLower(chartVersion), "-rc")
```

briandowns (changes requested):

pkg/lifecycle/lifecycle.go
rafaelbreno: There's a few other places that uses this error, and IMO we should declare every custom error, so when we unit test we have a variable to compare the expected error.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/charts-build-scripts
 
-go 1.18
```
tashima42, line 3: Shouldn't we just go ahead and upgrade to `1.23`?

pkg/lifecycle/version_rules.go
```diff
@@ -0,0 +1,195 @@
...
+
+	minor, err := strconv.Atoi(parts[1])
+	if err != nil {
+		// Handle error
+		return ""
+	}
+
+	minor += number
+	return fmt.Sprintf("%d.%d", major, minor)
+}
```
tashima42, line 195: I believe the semver package that we already use in other parts of our code can be used to handle this

pkg/lifecycle/lifecycle.go
nicholasSUSE: The error is used in 3 different places for now. 
 @briandowns 
 
It is a general one that will be thrown if you have done something really nasty to the charts repository local folder. 
That is why I am not doing 3 distinct errors and only a general one. 
@rafaelbreno

pkg/lifecycle/parser_test.go
nicholasSUSE: Fixed and also implemented assert package for unit-testing, it is easier to read. @briandowns

pkg/lifecycle/version_rules.go
nicholasSUSE: In this case I need it, otherwise I will get an error saying that `min` and `max` variable are declared and not used. 
@briandowns

pkg/lifecycle/version_rules.go
nicholasSUSE: I tried this before; it adds more complexity to a simple thing. 
In this case, I prefer to do the raw operations. 
@tashima42

tashima42: 👍 I just noticed this is not ecm distro tools

go.mod
nicholasSUSE: Just did it @tashima42

tashima42 (approved):

PR #150, opened 2024-09-11
Compare index.yaml for integrity of git pages. 

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/475

## Solution
We are implementing a command to load the current `index.yaml` from the charts repository, download the index.yaml from `charts.rancher.io`, and compare both. 

This will be used on every release Pull Request at: [charts-repo](https://github.com/rancher/charts)
Only on the last active branch.

### DIFF
--- a/pkg/auto/validate.go
+++ b/pkg/auto/validate.go
@@ -1,17 +1,26 @@
 package auto
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"os"
 	"strconv"
 
 	"github.com/Masterminds/semver"
+	"github.com/go-git/go-billy/v5"
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-github/v41/github"
+	"github.com/rancher/charts-build-scripts/pkg/helm"
 	"github.com/rancher/charts-build-scripts/pkg/lifecycle"
 	"github.com/rancher/charts-build-scripts/pkg/options"
 	"github.com/rancher/charts-build-scripts/pkg/path"
+	"github.com/rancher/charts-build-scripts/pkg/rest"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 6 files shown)


pkg/auto/validate.go
```diff
@@ -170,3 +179,45 @@ func (v *validation) checkNeverModifyReleasedChart(assetFilePaths map[string]str
 
 	return nil
 }
+
+// CompareIndexFiles will load the current index.yaml file from the root filesystem and compare it with the index.yaml file from charts.rancher.io
+func CompareIndexFiles(rootFs billy.Filesystem) error {
```
briandowns, line 184: This function returns an error in some cases and also exits in other cases. There should only be 1 entry and exit point for a program, generally. Can we update this to return an error consistently and have the caller made the decision to exit or otherwise?

pkg/rest/request.go
```diff
@@ -50,3 +51,18 @@ func Post(url string, body, responseModel any) error {
 
 	return nil
 }
+
+// Get sends a GET request to the given URL and returns the response body as a byte slice.
+func Get(url string) ([]byte, error) {
```
briandowns, line 56: This is somewhat of an anti pattern but ssince we're going to move away from this repo or refactor over time I'll ignore it but generally, unless there are enrichments to the request, there's no need to wrap the call. Also, it's perferred to use a request and set a timeout as this could spin indefinitely otherwise.

pkg/auto/validate.go
nicholasSUSE: Fixed

pkg/rest/request.go
nicholasSUSE: You are totally right, fixing it.

tashima42 (approved):

PR #151, opened 2024-10-07
Push Helm Chart to OCI Registry

nicholasSUSE:
This was made formerly for this issue: https://github.com/rancher/ecm-distro-tools/issues/476
But it was discontinued; however, I kept this PR, and it will serve for the `Prime-Charts` pushed to an OCI registry. 

Issue: https://github.com/rancher/ecm-distro-tools/issues/658

Implementing a CLI command to be called by a GHA job on every chart release that will push the released chart to an OCI registry.


main.go
```diff
@@ -836,3 +892,18 @@ func compareIndexFiles(c *cli.Context) {
 	}
 	fmt.Println("index.yaml files are the same at git repository and charts.rancher.io")
 }
+
+func updateOCIRegistry(c *cli.Context) {
+
+	if OciDNS == "" || OciUser == "" || OciPassword == "" {
+		fmt.Printf("OCI_DNS, OCI_USER, OCI_PASS environment variables must be set to run update-oci-registry\n")
```
tashima42, line 899:
```suggestion
		fmt.Println("OCI_DNS, OCI_USER, OCI_PASS environment variables must be set to run update-oci-registry")
```

main.go
```diff
@@ -836,3 +892,18 @@ func compareIndexFiles(c *cli.Context) {
...
+		os.Exit(1)
+	}
+
+	rootFs := filesystem.GetFilesystem(getRepoRoot())
+	if err := auto.UpdateOCI(rootFs, OciDNS, OciUser, OciPassword, DebugMode); err != nil {
+		fmt.Printf("failed to update oci registry: %v \n", err)
```
tashima42, line 905:
```suggestion
		fmt.Println("failed to update oci registry: ", err.Error())
```

pkg/auto/oci.go
```diff
@@ -0,0 +1,192 @@
...
+			if exists {
+				fmt.Printf("Asset %s already exists in the OCI registry \n", asset)
+				return pushedAssets, fmt.Errorf("asset %s already exists in the OCI registry", asset)
+			}
+
+			fmt.Printf("Pushing asset to OCI Registry: %s \n", asset)
+			if err := o.push(o.helmClient, assetData, buildPushURL(o.DNS, chart, version)); err != nil {
+				err = fmt.Errorf("failed to push %s; error: %w ", asset, err)
+				fmt.Printf(err.Error())
+				return pushedAssets, err
```
tashima42, line 148: It would be better if we could make an atomic transaction and push all assets or push none, but if this isn't possible, you should do the same as you're doing in the last err check block, composing the errors instead of just logging them

tashima42 (approved):

PR #153, opened 2024-11-14
Slsa charts bypass

nicholasSUSE:
Long story short:

Charts that have images supporting SLSA must no longer be synced from:

docker hub -> primer registry. 

### Solution

On the charts repository, there will be a new file called `slsa.yaml`; see: https://github.com/rancher/charts/pull/4760

The chart owners will update this file with the new slsa compliant images. 
We bypass the `regsync process` for the prime registry for all these images.

### DIFF
--- a/pkg/path/path.go
+++ b/pkg/path/path.go
@@ -6,41 +6,54 @@ const (
 
 	// RepositoryHelmIndexFile is the file on your Staging/Live branch that contains your Helm repository index
 	RepositoryHelmIndexFile = "index.yaml"
+
 	// RepositoryPackagesDir is a directory on your Staging branch that contains the files necessary to generate your package
 	RepositoryPackagesDir = "packages"
+
 	// RepositoryAssetsDir is a directory on your Staging/Live branch that contains chart archives for each version of your package
 	RepositoryAssetsDir = "assets"
+
 	// RepositoryChartsDir is a directory on your Staging/Live branch that contains unarchived charts for each version of your package
 	RepositoryChartsDir = "charts"
 
 	// PackageOptionsFile is the name of a file that contains information about how to prepare your package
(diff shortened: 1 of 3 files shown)


pkg/regsync/generateconfig.go
```diff
@@ -217,3 +220,44 @@ func walkMap(inputMap interface{}, callback func(map[interface{}]interface{})) {
...
+	}
+	return nil
+}
+
+func readSlsaYaml() ([]string, error) {
+	var slsaImgs []string
+
+	file, err := os.Open(path.SlsaYamlFile)
+	if err != nil {
+		return nil, nil // backward version compatibility
```
tashima42, line 250: Shouldn't this be an error?

pkg/path/path.go
```diff
@@ -50,12 +63,15 @@ const (
 	// RepositoryLogosDir is a directory on your Staging/Live branch that contains the files with the logos of each chart
 	RepositoryLogosDir = "assets/logos"
 
-	// RepositoryStAte file is a file to hold the current status of the released and developed assets versions
+	// RepositoryStateFile file is a file to hold the current status of the released and developed assets versions
```
pjbgf, line 66:
```suggestion
	// RepositoryStateFile is a file to hold the current status of the released and developed assets versions
```

pkg/regsync/generateconfig.go
pjbgf, line 250:
```suggestion
		return nil, err // backward version compatibility
```

pjbgf (changes requested):

pkg/regsync/generateconfig.go
nicholasSUSE: This feature should only be available for release-v2.10 and up. 
There will be no slsa.yaml files for the older branches. 
So this should not block the code execution.

PR #157, opened 2024-12-11
changing token strategy

nicholasSUSE:
Token strategy for docker hub authentication has changed

### DIFF
--- a/pkg/images/checkImages.go
+++ b/pkg/images/checkImages.go
@@ -136,16 +136,16 @@ func retrieveToken() (string, error) {
 // retrieveCredentials retrieves credentials from environment variables
 func retrieveCredentials() *TokenRequest {
 
-	username := os.Getenv("DOCKER_USERNAME")
-	password := os.Getenv("DOCKER_PASSWORD")
+	username := os.Getenv("DOCKERHUB_USERNAME")
+	password := os.Getenv("DOCKERHUB_TOKEN")
 
 	if strings.Compare(username, "") == 0 {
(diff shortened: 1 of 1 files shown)


pkg/images/checkImages.go
```diff
@@ -40,10 +40,7 @@ func CheckImages() error {
 	imagesOutsideNamespace := checkPattern(imageTagMap)
 
 	// Get a token to access the Docker Hub API
-	token, err := retrieveToken()
-	if err != nil {
-		logrus.Infof("failed to retrieve token, requests will be unauthenticated: %v", err)
-	}
+	token := getTokenFromEIO()
```
tashima42, line 43: I don't think this would work, the token still needs to be generated using the credentials

nicholasSUSE: I was assuming:
```
          secrets: |
            secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials username | DOCKERHUB_USERNAME;
            secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials password | DOCKERHUB_TOKEN;
```

This would get the token since it is called dockerhub_token. 
Maybe that is just the password and we still have to get the token right? 

@tashima42

tashima42: Yes, it's just the password

nicholasSUSE: Ok thanks!

tashima42:
I think it would be nice to change these variable names on the charts PR instead of here to maintain compatibility with existing pipelines, it's easy enough to change the variable names there

Issue #130, opened 2024-04-02
Document requirement for GNU Diff/Patch


mallardduck:
The functionality of generating and applying the diff patches are dependent on local system `diff` and `patch` binaries. 

Per these lines:

https://github.com/rancher/charts-build-scripts/blob/a7001e331fda7b61c8df358862f46436eaae9df2/pkg/diff/diff.go#L19
https://github.com/rancher/charts-build-scripts/blob/a7001e331fda7b61c8df358862f46436eaae9df2/pkg/diff/diff.go#L57

Unfortunately, a "normal assumption" would probably be that "This is all a golang script, so the execution shouldn't be affected by local environment variances." So without this requirement documented clearly then we can encounter unexpected behavior explained by "well it works on my machine".

This was all learned in a first hand experience based on PRs I was working on last friday. We were seeing that even if I made no changes at all and ran `make charts` then it would results in file modifications. Eventually it was all tracked down to the fact that on macOS they ship with a BSD based "Apple" version of `diff` and `patch`. See:

```
→ patch --version
patch 2.0-12u11-Apple

→ diff --version
Apple diff (based on FreeBSD diff)
```

After using brew to install `gpatch` and `diffutils` packages I have GNU versions and get expected results:

```
→ patch --version
GNU patch 2.7.6

→ diff --version
diff (GNU diffutils) 3.10
```

nicholasSUSE:
@mallardduck 

Please, this is the documentation I am providing to workaround this problem:
https://github.com/rancher/charts/wiki/0‐Chart-Owners#dependencies

Feel free to give a review. 


I am also providing the check as code: https://github.com/rancher/charts-build-scripts/pull/163

PR #163, opened 2025-01-29
adding check for apple free-BSD dependencies

nicholasSUSE:
Issue: https://github.com/rancher/charts-build-scripts/issues/130

At the beginning of this semester, we are moving to auto-chart-bumps, which will be performed in GHA actions. 

Therefore, I am implementing this check and improving documentation so that developers using Darwin machines can install the dependencies properly. 

Documentation: https://github.com/rancher/charts/wiki/0‐Chart-Owners#dependencies

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -3,13 +3,31 @@ package diff
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
 	"github.com/sirupsen/logrus"
 )
+
+func checkDependencyPackage(pathToCmd string) error {
+	cmd := exec.Command(pathToCmd, "--version")
+	out, err := cmd.Output()
+	if err != nil {
+		return err
+	}
+
+	version := string(out)
+
+	if strings.Contains(version, "Apple") || strings.Contains(version, "FreeBSD") {
+		return errors.New("detected Apple/FreeBSD version of patch. This will lead to compatibility issues. Install GNU patch: https://github.com/rancher/charts-build-scripts/issues/130")
+	}
+
+	return nil
+}
 
 // GeneratePatch generates the patch between the files at srcPath and dstPath and outputs it to patchPath
 // It returns whether the patch was generated or any errors that were encountered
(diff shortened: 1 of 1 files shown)


pkg/diff/diff.go
```diff
@@ -3,14 +3,32 @@ package diff
...
+	cmd := exec.Command(pathToCmd, "--version")
+	out, err := cmd.Output()
+	if err != nil {
+		return err
+	}
+
+	version := string(out)
+
+	if strings.Contains(version, "Apple") || strings.Contains(version, "FreeBSD") {
+		return errors.New("detected Apple/FreeBSD version of patch. This will lead to compatibility issues. Install GNU patch: https://github.com/rancher/charts-build-scripts/issues/130")
```
apoorvajagtap, line 26: As this check is not specific to `patch` command, should we modify the error message something similar to? 
```suggestion
		return errors.Errorf("detected Apple/FreeBSD version of %[1]s. This will lead to compatibility issues. Install GNU %[1]s: https://github.com/rancher/charts-build-scripts/issues/130", pathToCmd)
```

briandowns (approved):

apoorvajagtap (approved):
rest LGTM

pkg/diff/diff.go
```diff
@@ -3,14 +3,32 @@ package diff
 import (
 	"bufio"
 	"bytes"
+	"errors"
```
apoorvajagtap, line 6:
```suggestion
	"github.com/pkg/errors"
```
Looking the test failures, I realised that we need to import a different package. `errors` doesn't have `Errorf` function.

apoorvajagtap (changes requested):

pkg/diff/diff.go
mallardduck: Rather than this - I believe a more apt solution will be: `fmt.Errorf`. 😄

pkg/diff/diff.go
```diff
@@ -3,14 +3,32 @@ package diff
...
+	cmd := exec.Command(pathToCmd, "--version")
+	out, err := cmd.Output()
+	if err != nil {
+		return err
+	}
+
+	version := string(out)
+
+	if strings.Contains(version, "Apple") || strings.Contains(version, "FreeBSD") {
+		return errors.Errorf("detected Apple/FreeBSD version of %[1]s. This will lead to compatibility issues. Install GNU %[1]s: https://github.com/rancher/charts-build-scripts/issues/130", pathToCmd)
```
mallardduck, line 26:
```suggestion
		return fmt.Errorf("detected Apple/FreeBSD version of %[1]s. This will lead to compatibility issues. Install GNU %[1]s: https://github.com/rancher/charts-build-scripts/issues/130", pathToCmd)
```

mallardduck (changes requested):

mallardduck (approved):
👍 LGTM

PR #165, opened 2025-02-04
Add a "soft-errors" mode for prepare subcommand

mallardduck:
This is useful in certain rebase actions teams with complicated charts may face. Essentially this flag will allow for all existing patches to be applied when preparing even if they would cause errors. This way someone doing a rebase can see all the conflicting patches in one execution.

They can either try to fix as many as possible, or opt to temporarily disable them and recreate them.

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/rancher/charts-build-scripts/pkg/util"
 	"io/ioutil"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -253,7 +265,7 @@ func main() {
 			Usage:  "Pull in the chart specified from upstream to the charts directory and apply any patch files",
 			Action: prepareCharts,
 			Before: setupCache,
-			Flags:  []cli.Flag{packageFlag, cacheFlag},
+			Flags:  []cli.Flag{packageFlag, cacheFlag, softErrorsFlag},
```
nicholasSUSE, line 268: I think you will need to `softErrorsFlag` also to make charts Flags (line 282). 

`make charts` command also calls `make prepare` internally before running. 

You will have the same problem later on that stage.

mallardduck: For now I've excluded `charts` subcommand intentionally - I'm open to expanding there, I just wanted to put some more consideration into if there should be more DX features around that one.

For now I'd expect - at least on O&B team - we would do something like:

- Find all patches with errors,
- Move those to a `patch-off` folder from `patch`,
- Be able to run `make chart`
- Work to recreate/fix the patches that were disabled

nicholasSUSE: Ok approved.

nicholasSUSE (approved):

PR #186, opened 2025-02-13
feat(dev) : specify root repo path using DEV_REPO_ROOT for debugging

alexandreLamarre:
- Allows specifying relative/absolute root repo paths for debugging by developers

- correctly pass in configuration.yaml path config flag : places where `defaultChartsScriptOptionsFile` were used should've been using `ChartsScriptOptionsFile` instead
```go
configFlag := cli.StringFlag{
		Name:        "config",
		Usage:       "A configuration file with additional options for allowing this branch to interact with other branches",
		TakesFile:   true,
		Destination: &ChartsScriptOptionsFile,
		Value:       defaultChartsScriptOptionsFile,
	}
```

- Migrated references of deprecated ioutil package to stable:
  - `ioutil.ReadFile` -> `os.Readfile`
  - `ioutil.WriteFile` -> `os.WriteFile`
  - `ioutil.MkdirTemp` -> `os.Mkdirtemp`

### DIFF
--- a/pkg/charts/checkRCCharts.go
+++ b/pkg/charts/checkRCCharts.go
@@ -1,7 +1,6 @@
 package charts
 
 import (
-	"os"
 	"strings"
 
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
@@ -10,14 +9,7 @@ import (
 )
 
 // CheckRCCharts checks for any charts that have RC versions
-func CheckRCCharts() map[string][]string {
-
-	// Get the current working directory
-	repoRoot, err := os.Getwd()
-	if err != nil {
-		logrus.Fatalf("Unable to get current working directory: %s", err)
-	}
-
+func CheckRCCharts(repoRoot string) map[string][]string {
 	// Get the filesystem on the repo root
 	repoFs := filesystem.GetFilesystem(repoRoot)
 
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -2,7 +2,6 @@ package lifecycle
 
 import (
 	"errors"
-	"os"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/rancher/charts-build-scripts/pkg/filesystem"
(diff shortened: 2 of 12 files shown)


pkg/charts/checkRCCharts.go
```diff
@@ -10,14 +9,7 @@ import (
...
 // CheckRCCharts checks for any charts that have RC versions
-func CheckRCCharts() map[string][]string {
-
-	// Get the current working directory
-	repoRoot, err := os.Getwd()
-	if err != nil {
-		logrus.Fatalf("Unable to get current working directory: %s", err)
-	}
-
+func CheckRCCharts(repoRoot string) map[string][]string {
```
alexandreLamarre, line 12: One of the main reasons, I've removed os.Getwd() definitions in package functions is to get rid of scenarios where all functions are not sharing the same abstraction for the repoRoot -- in the case where we make changes like this one.

In this PR I'm preferring using getRepoRoot() function defined in main.go and passing it through to sub functions as that seemed to be mostly the standard when Arvind was still working on the project

pkg/lifecycle/lifecycle.go
```diff
@@ -37,13 +36,10 @@ type WalkDirFunc func(fs billy.Filesystem, dirPath string, doFunc filesystem.Rel
...
 // If anything fails the operation will be aborted.
-func InitDependencies(rootFs billy.Filesystem, branchVersion string, currentChart string) (*Dependencies, error) {
+func InitDependencies(repoRoot string, rootFs billy.Filesystem, branchVersion string, currentChart string) (*Dependencies, error) {
 	var err error
 
-	workDir, err := os.Getwd()
-	if err != nil {
-		return nil, err
-	}
+	workDir := repoRoot
```
alexandreLamarre, line 42: Could alternatively pass in just the repoRoot string and construct the billy.Filesystem inside this function to avoid duplicate parameters in function signatures, although I was ensure if the workDir should ALWAYS be the repoRoot in the autobumps/lifecycle code

pkg/charts/checkRCCharts.go
alexandreLamarre: We could also have a shared variable in pkg/charts, along the lines of:

```go
func init() {

  repoRoot, err := os.Getwd()
  if err != nil {
    logrus.Fatal(err)
  } 
  DefaultRepoRoot= repoRoot
}

var (
  // default value
  DefaultRepoRoot = ""
  // flag writes to this one
  RepoRoot = ""
)
```

which could get passed in / written to as a cli.Flag, and consumers would use as `charts.RepoRoot`, but I opted for simplicity here

alexandreLamarre:
Note this also allows for embedding git submodules which would be useful for integration tests in `testify`, which I think would both greatly reduce the amount of bugs shipped to charts-build-scripts, but also increase iteration speed

nicholasSUSE (approved):
LGTM: 

Please document the use of this on `charts` wiki or in this PR so I can update the documentation there. 

Thank you for your work

alexandreLamarre:
Updated documentation : https://github.com/rancher/charts-build-scripts/wiki/Debugging-charts%E2%80%90build%E2%80%90scripts

PR #204, opened 2025-04-07
major refactor: logging, configuration and auto-bump trackability

nicholasSUSE:
Major improvements in: 
- Logging with `slog` + custom wrapper 
- Getting repo root of `charts repository` execution
- Logging result of auto-bump to `config/bump.json`
- re-organizing configuration and debug files to separate it from Chart related files 

After the release 1.4.9, there is a problem: `make validate` works perfectly at local machines; however, in GHA, it breaks, but the logs do not show the reason.


pkg/util/log.go
```diff
@@ -0,0 +1,38 @@
+package util
```
briandowns, line 1: Rename this package log or logging. util doesn't say anything about what's in it and it's currently only log functions.

pkg/util/log.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"log/slog"
+	"os"
+	"runtime"
+	"time"
+)
+
+func Log(lvl slog.Level, msg string, attrs ...slog.Attr) {
+	// internal context because it is a limited CLI app
+	ctx := context.Background()
```
briandowns, line 13: This function should take a context created when the app starts in main and passed down and around. Creating one for each call is bad practice and not good for performance.

pkg/util/log.go
```diff
@@ -0,0 +1,38 @@
...
+	// Caller information (PC, Func, etc)
+	var pcs [1]uintptr
+	runtime.Callers(2, pcs[:])
+	fs := runtime.CallersFrames(pcs[:])
+	f, _ := fs.Next()
+
+	record := slog.NewRecord(time.Now(), lvl, msg, f.PC)
+	// Add any attributes passed to the Log function to the record
+	record.AddAttrs(attrs...)
+	_ = logger.Handler().Handle(ctx, record)
```
briandowns, line 27: You can remove the underscore and equal sign.

pkg/util/log.go
nicholasSUSE: By passing down and around, do you mean passing across functions or setting it as a global variable?

briandowns: Creating in the main function and passing it around to any function that needs it.

nicholasSUSE: done

briandowns (approved):

PR #214, opened 2025-04-11
check upstream app version

nicholasSUSE:
Add a safety check for when the chart has already been bumped.

### DIFF
--- a/pkg/auto/chart_bump.go
+++ b/pkg/auto/chart_bump.go
@@ -245,6 +245,18 @@ func (b *Bump) BumpChart(ctx context.Context, versionOverride string) error {
 		return err
 	}
 
+	// check if the version to bump does not already exists
+	alreadyExist, err := checkBumpAppVersion(ctx, b.Pkg.UpstreamChartVersion, b.assetsVersionsMap[b.targetChart])
+	if err != nil {
+		logger.Log(ctx, slog.LevelError, "error while checking if version exists", logger.Err(err))
+		return err
+	}
+	if alreadyExist {
+		logger.Log(ctx, slog.LevelError, "version to bump already exists", slog.String("version", *b.Pkg.UpstreamChartVersion))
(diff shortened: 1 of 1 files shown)


pkg/auto/chart_bump.go
```diff
@@ -245,6 +245,18 @@ func (b *Bump) BumpChart(ctx context.Context, versionOverride string) error {
 		return err
 	}
 
+	// check if the version to bump does not already exists
+	alreadyExist, err := checkBumpAppVersion(ctx, b.Pkg.UpstreamChartVersion, b.assetsVersionsMap[b.targetChart])
+	if err != nil {
+		logger.Log(ctx, slog.LevelError, "error while checking if version exists", logger.Err(err))
```
briandowns, line 251: For future reference only, you probably want to avoid logging an error and returning that error because it opens up the possibility of logging it twice in the throwing function and the originally called function.

pkg/auto/chart_bump.go
```diff
@@ -245,6 +245,18 @@ func (b *Bump) BumpChart(ctx context.Context, versionOverride string) error {
...
+		return err
+	}
+	if alreadyExist {
+		logger.Log(ctx, slog.LevelError, "version to bump already exists", slog.String("version", *b.Pkg.UpstreamChartVersion))
+		git.FullReset() // quitting the job regardless if this works or not
+		return fmt.Errorf("version to bump already exists: %s", *b.Pkg.UpstreamChartVersion)
```
briandowns, line 257: Generally, you want to avoid any format functions when using strings. They can be concatenation with the "+" operator and avoid all of the unnecessary reflection and memory use.

briandowns (approved):

pkg/auto/chart_bump.go
nicholasSUSE: You are right here; since this function is only called at BumpChart, it is unnecessary to log it. 

I usually log the error in functions that are used by several others.

PR #216, opened 2025-05-08
Validate icons are downloaded

nicholasSUSE:
In order to prevent this https://github.com/rancher/ecm-distro-tools/issues/553 from happening again. 

This check will ensure that every new chart has downloaded the icons file locally in the repository. 

If the chart owner has not downloaded, the CI will block that PR from merging. 

A few Charts can bypass this rule because they are auto-installable and have no icon to present in the UI.

### DIFF
--- a/pkg/auto/release.go
+++ b/pkg/auto/release.go
@@ -197,7 +197,7 @@ func (r *Release) PullIcon(ctx context.Context, rootFs billy.Filesystem) error {
 	// Check if the icon exists in the dev branch
 	if err := r.git.CheckFileExists(relativeIconPath, r.VR.DevBranch); err != nil {
 		logger.Log(ctx, slog.LevelError, "icon file not found in dev branch but should", slog.String("icon", relativeIconPath), logger.Err(err))
-		return fmt.Errorf("icon file not found in dev branch but should: %w", err)
+		return errors.New("icon file not found in dev branch but should: " + err.Error())
 	}
 
 	// checkout the icon file from the dev branch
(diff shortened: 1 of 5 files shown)


pkg/auto/release.go
```diff
@@ -211,3 +208,17 @@ func (r *Release) PullIcon(ctx context.Context, rootFs billy.Filesystem) error {
...
+
+func loadChartYaml(rootFs billy.Filesystem, chart string, chartVersion string) (*chart.Metadata, error) {
+	// Get Chart.yaml path and load it
+	chartYamlPath := path.RepositoryChartsDir + "/" + chart + "/" + chartVersion + "/Chart.yaml"
+	absChartPath := filesystem.GetAbsPath(rootFs, chartYamlPath)
+
+	// Load Chart.yaml file
+	chartMetadata, err := helmChartutil.LoadChartfile(absChartPath)
+	if err != nil {
+		return nil, fmt.Errorf("could not load %s: %s", chartYamlPath, err)
```
rafaelbreno, line 220: I would suggest to change the `%s` to `%v` for the `err` value.
```suggestion
		return nil, fmt.Errorf("could not load %s: %v", chartYamlPath, err)
```

briandowns (approved):

rafaelbreno (approved):
left a comment, but not a blocker.

pkg/auto/release.go
nicholasSUSE: Implemented in a comment what you asked, thank you for the heads-up

PR #220, opened 2025-05-16
Update version handling to support more prerelease types

thardeck:
It matches **RC**, **Alpha**, and **Beta** prerelease versions with optional hyphens/dots and optional version numbers (e.g., `-rc1`, `.alpha2`, `beta`, etc.), case-insensitively.

Before only release candidates were supported.


pkg/auto/chart_bump.go
```diff
@@ -336,19 +336,21 @@ func (b *Bump) BumpChart(ctx context.Context, versionOverride string, multiRCs b
 	}
 
 	if !multiRCs {
-		if strings.Contains(b.versions.toRelease.txt, "-rc") {
+		if strings.Contains(strings.ToLower(b.versions.toRelease.txt), "rc") ||
```
nicholasSUSE, line 339: ok this is a problem, there is a high chance that some chart name in some situation might have a "rc" on its name. 

We can do here:
```
-rc || .rc
```

pkg/auto/chart_bump.go
```diff
@@ -336,19 +336,21 @@ func (b *Bump) BumpChart(ctx context.Context, versionOverride string, multiRCs b
...
+			strings.Contains(strings.ToLower(b.versions.toRelease.txt), "alpha") ||
```
nicholasSUSE, line 340: same thing for "alpha" and "beta" 

Something like: 
```
-alpha || .alpha
-beta || .beta
```

pkg/auto/chart_bump.go
```diff
@@ -547,19 +549,41 @@ func checkBumpAppVersion(ctx context.Context, bumpAppVersion *string, versions [
 	return false, nil
 }
 
-func listRCVersions(rcVersion string, assets []lifecycle.Asset) ([]string, error) {
-	idx := strings.Index(rcVersion, "-rc")
-	if idx == -1 {
-		return nil, fmt.Errorf("invalid rcVersion format: %s", rcVersion)
+func listPrereleaseVersions(version string, assets []lifecycle.Asset) ([]string, error) {
+	prereleaseTypes := []string{"rc", "alpha", "beta"}
```
nicholasSUSE, line 553: Change this logic accordingly, to the changes above, please:

``` 
	prereleaseTypes := []string{"-rc",".rc", "-alpha", ".alpha", "-beta", ".beta"}
```

nicholasSUSE (changes requested):
@thardeck 

Thank you for your PR. 
I see that you also added the unit-tests thank you very much for that. 

I asked for a few changes. 

Please provide a good description just with why and what has changed.

PR #237, opened 2025-07-17
Icons new safety check

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/591

checking if the file pointed to by the Chart.yaml also exists.

### DIFF
--- a/pkg/auto/validate.go
+++ b/pkg/auto/validate.go
@@ -330,7 +330,7 @@ func loadAndCheckIconPrefix(ctx context.Context, rootFs billy.Filesystem, chart 
 	}
 
 	if !exist {
-		return errors.New("icon path is a file:// prefix, but the icon does not exist, after make prepare, you need to run make icon for chart:" + chart)
+		return errors.New("icon path is a file:// prefix, but the icon does not exist, after 'make prepare', you need to run 'make icon' for chart:" + chart)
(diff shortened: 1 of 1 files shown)


briandowns (approved):

pkg/auto/validate.go
```diff
@@ -323,5 +324,14 @@ func loadAndCheckIconPrefix(ctx context.Context, rootFs billy.Filesystem, chart
 		return errors.New("icon path is not a file:// prefix, after make prepare, you need to run make icon for chart:" + chart)
 	}
 
+	exist, err := filesystem.PathExists(ctx, rootFs, strings.TrimPrefix(iconField, "file://"))
+	if err != nil {
+		return err
+	}
+
+	if !exist {
+		return errors.New("icon path is a file:// prefix, but the icon does not exist, after make prepare, you need to run make icon for chart:" + chart)
```
rafaelbreno, line 333: What do you think about adding single quotes to the `make` commands? 
```suggestion
		return errors.New("icon path is a file:// prefix, but the icon does not exist, after 'make prepare', you need to run 'make icon' for chart:" + chart)
```

rafaelbreno (approved):
Just a single comment, not a blocker. LGTM

PR #377, opened 2026-07-03
check if chart is using latest non-major image version

diogoasouza:
This PR adds the command `validate-image-versions`

### What does it do?

The command checks whether the images used by a chart are on their latest available minor/patch version, so that outdated images can be caught before release rather than during manual review.

## How it works
   - Receives a chart and the version as inputs.
   - Fetches the list of images to check from `config/image-version-check.yaml` on the `automation-core` branch.
   - Gets all images used by the chart.
   - For each image found in the chart, it pulls the available tags from Docker Hub (for now) and compares them against the chart's current tag.
   - Prints a JSON `Report` with per-image results (`currentTag`, `latestAvailable`, `needsUpdate`), plus any configured images that weren't found in the chart.
   
## Usage

   ```
   charts-build-scripts validate-image-versions --chart <chart> --chartVersion <version>
   ```

### Output example

  ```
  {"chart":"rancher-backup","needsUpdate":true,"images":[{"repository":"rancher/kuberlr-kubectl","currentTag":"v8.0.0","latestAvailable":"v8.1.1","needsUpdate":true}]}
  ```


image-version-check.yaml
```diff
@@ -0,0 +1,3 @@
+images:
```
nicholasSUSE: We have a new canonical source of truth for configuration files in charts. 

See: https://github.com/rancher/charts/blob/automation-core/config/blocklist.yaml
Implementation PR: https://github.com/rancher/charts-build-scripts/pull/364

There are some reusable functions for you in the PR above. 

Also, please commit more often, in small chunks of code.

nicholasSUSE (changes requested):

briandowns (approved):

pkg/filesystem/assets.go
```diff
@@ -118,6 +118,22 @@ func DecodeValueYamlInTgz(ctx context.Context, tgzPath string, fileNames []strin
 	}
 }
 
+// DecodeValuesYamlFile reads a values.yaml file from disk and returns a normalized map.
+// Keys are coerced to strings and numeric values are stringified for consistent lookups.
+func DecodeValuesYamlFile(ctx context.Context, filePath string) (map[string]interface{}, error) {
```
nicholasSUSE, line 123: `DecodeValuesYamlFile` duplicates file-opening logic already in `LoadYamlFile` at `pkg/filesystem/yaml.go.

Try: 
```
  func DecodeValuesYamlFile(ctx context.Context, filePath string) (map[string]interface{}, error) {
      values, err := LoadYamlFile[map[string]interface{}](ctx, filePath, true)
      if err != nil {
          return nil, fmt.Errorf("decoding values file %q: %w", filePath, err)
      }
      if values == nil {
          return map[string]interface{}{}, nil
      }
      return normalizeMapStructure(*values).(map[string]interface{}), nil
  }
```

Also you are using yaml.v2, although I maintain support to it because of older charts, we should not use it and we should always aim for safe decoding (which ensures yaml files are following the yaml standard).

main.go
```diff
@@ -414,6 +417,13 @@ func main() {
 		Destination: &IsPrimeChart,
 		EnvVar:      defaultIsPrimeChartVariable,
 	}
+	imageVersionConfigFlag := cli.StringFlag{
```
nicholasSUSE: path.ImageVersionCheckFile should not be passed as a cli flag 

It is a global configuration value. 

The function should call it directly. 

Please remove this flag.

pkg/registries/imageversion.go
```diff
@@ -0,0 +1,162 @@
...
+	SkippedUnsupported []string      `json:"skippedUnsupported,omitempty"`
+}
+
+var LoadImageVersionList = config.LoadImageVersionList
+
+// latestSameMajor returns the highest available tag that shares the same major version
+// as current and has no pre-release suffix, along with whether an update is needed.
+// Non-semver and pre-release tags in available are skipped.
+// Returns ("", false) when current cannot be parsed as semver.
+func latestSameMajor(current string, available []string) (string, bool) {
```
nicholasSUSE: This is a great, useful function, 
I am starting to centralize all version utilities functions at: 'pkg/util/versions.go`

Once/if it grows enough, it will have its own separate pkg. 

Please move it there. 


---

> Context: The semver/v3 has its version pinned in the dependencies because after the pinned version, it doesn't recognize our versioning standard on rancher-charts (x.y.z+up.ab.c) as a valid semver. This is a huge pain that I still need to resolve, and if I have all version methods and helpers on a centralized place, it will be easier once I have to deal with.... (estimated time to deal with this is end of august once I will have to update the helm dependency to v4)

pkg/config/imageVersion.go
```diff
@@ -0,0 +1,48 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// ImageVersionCheckOptions holds the list of images to check for version updates.
+type ImageVersionCheckOptions struct {
+	Images []ImageVersionEntry `yaml:"images"`
+}
+
+// ImageVersionEntry describes a single image to validate.
+type ImageVersionEntry struct {
```
nicholasSUSE: Please, let's change this to: 
``` 
  type ImageConfig struct {
      Repository string `yaml:"repository"`
      Tag        string `yaml:"tag,omitempty"`  // optional
  }
```

pkg/config/imageVersion.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"errors"
+
+	"github.com/rancher/charts-build-scripts/pkg/git"
+	"github.com/rancher/charts-build-scripts/pkg/path"
+	"gopkg.in/yaml.v2"
+)
+
+// ImageVersionCheckOptions holds the list of images to check for version updates.
+type ImageVersionCheckOptions struct {
```
nicholasSUSE: Let's adopt the following pattern at https://github.com/rancher/charts/pull/8093/changes

```
<image>:
     repository: <repository> 
     tag: <tag> 
``` 

In such case your example would be: 
```
kuberlr: 
     repository: rancher/kuberlr-kubectl
     tag: ""
``` 

Also, please at to the top of that file a comprehensive comment explanation of what it is.

pkg/config/imageVersion.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	// Fetch image-version-check.yaml from automation-core branch
+	data, err := repo.ShowFileFromRemoteBranch(ctx, path.AutoCoreBranch, path.ImageVersionCheckFile)
+	if err != nil {
+		return nil, errors.New("load image-version-check show: " + err.Error())
+	}
+
+	// Parse YAML directly into struct
+	var opts ImageVersionCheckOptions
```
nicholasSUSE, line 40: With the above fixes, we would have here: 
``` 
      var imageMap map[string]ImageConfig
      if err := yaml.Unmarshal(data, &imageMap); err != nil {
          return nil, fmt.Errorf("load image-version-check unmarshal: %w", err)
      }
``` 

Pros: 
- images in the root level of a yaml file have to be unique by yaml standard
- O(1) in map searches.

pkg/registries/imageversion.go
```diff
@@ -0,0 +1,162 @@
...
+	if bestVer == nil {
+		return "", false
+	}
+	return bestTag, bestVer.GreaterThan(currentVer)
+}
+
+// collectChartImages walks the unpacked chart source under <repoRoot>/charts/<chart>/<version>
+// and returns a map of repository → []tags found in any values.yaml or values.yml.
+func collectChartImages(ctx context.Context, repoRoot, chart, version string) (map[string][]string, error) {
+	chartsBase := filepath.Join(repoRoot, "charts", chart)
```
nicholasSUSE: Try using the already created helper methods: 

``` 
  chartsBase := filepath.Join("charts", chart, version)

  exists, err := filesystem.PathExists(ctx, filesystem.GetFilesystem(repoRoot), chartsBase)
  if err != nil || !exists {
      return nil, errors.New("chart directory not found: " +  chartsBase)
  }
```

pkg/registries/imageversion.go
```diff
@@ -0,0 +1,162 @@
...
+	if _, err := os.Stat(chartsBase); os.IsNotExist(err) {
+		return nil, fmt.Errorf("chart directory not found: %s", chartsBase)
+	}
+
+	versionDir := filepath.Join(chartsBase, version)
+
+	repoTagMap := make(map[string][]string)
+	if err := walkValuesFiles(ctx, versionDir, repoTagMap); err != nil {
```
nicholasSUSE: No need for an entire new walk function, can use: 

```
  err = filesystem.WalkDir(ctx, fs, chartsBase, func(ctx context.Context, fs billy.Filesystem, path string, isDir bool) error {
      if isDir {
          return nil
      }
     ....
      // decode and traverse
  })
```

image-version-check.yaml
```diff
@@ -0,0 +1,3 @@
...
+  - name: kuberlr
```
nicholasSUSE: This needs to be removed.

nicholasSUSE (changes requested):

nicholasSUSE (approved):

PR #379, opened 2026-07-08
fix: Gracefully ignore blocklist outside expected Rancher repos

mallardduck:
The new `blocklist` feature for index generation is causing errors in the ORBS team repo where the `automation-core` branch does not exist. As a "simple" fix we can emit a custom error from `getUpstreamRemote` where the error ORBS repo triggers comes from.

Then subsequently inside of `LoadBlockList` we can use `errors.Is` to verify that this is the error we got so that the blocklist returns empty instead of an error which will gracefully fallback to the behavior that existed before `blocklist` was added.

### DIFF
--- a/pkg/config/blocklist.go
+++ b/pkg/config/blocklist.go
@@ -3,9 +3,11 @@ package config
 import (
 	"context"
 	"errors"
+	"log/slog"
 	"slices"
 
 	"github.com/rancher/charts-build-scripts/pkg/git"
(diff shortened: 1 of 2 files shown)


nicholasSUSE (approved):

nicholasSUSE:
Please add some context in the PR body.

briandowns (approved):

rmoioliveira (approved):

PR #385, opened 2026-07-22
Replace gopkg.in/yaml.vN with go.yaml.in/yaml/vN and github.com/ghodss/yaml with sigs.k8s.io/yaml

tashima42:
This PR replaces `gopkg.in/yaml.vN` with `go.yaml.in/yaml/vN` (preserving major version) and `github.com/ghodss/yaml` with `sigs.k8s.io/yaml` across the codebase.

Issue: https://github.com/rancher/rancher/issues/56157


nicholasSUSE:
@tashima42, before moving with this, let me add tests to make sure the migration won't break anything. 

Also, let's do this right after the release.

nicholasSUSE (approved):

PR #390, opened 2026-08-07
return trivy stderr on failure

diogoasouza:
Show trivy error on failure


pkg/registries/cve_check.go
```diff
@@ -1,10 +1,12 @@
 package registries
```
briandowns, line 1: NIT: Go's standard convention for file naming is to use underscores rather than camel case.

briandowns (approved):

pkg/registries/cve_check.go
diogoasouza: Honestly no idea why I named it like that lol

PR #394, opened 2026-08-10
skip image if it does not support the runner's platform

diogoasouza:
Some images are Windows-only, and since we run the CI on Linux, it fails

### DIFF
--- a/pkg/registries/cve_check.go
+++ b/pkg/registries/cve_check.go
@@ -15,7 +15,7 @@ import (
 )
 
 // errUnsupportedPlatform indicates trivy could not find an image for the runner's platform (Linux).
-var errUnsupportedPlatform = errors.New("image doesn't support Linux/amd64")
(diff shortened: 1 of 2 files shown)


briandowns (approved):

rmoioliveira (approved):

mallardduck (approved):

pkg/registries/cve_check.go
```diff
@@ -13,6 +14,12 @@ import (
 	"github.com/rancher/charts-build-scripts/pkg/helm"
 )
 
+// errUnsupportedPlatform indicates trivy could not find an image for the runner's platform (Linux).
+var errUnsupportedPlatform = errors.New("image doesn't support Linux/amd64")
```
tashima42: This might not always run on linux/amd64, we also have builds for linux/arm64 and darwin, it's better to remove the exact platform from the error message

diogoasouza: Done

tashima42 (approved):

PR #395, opened 2026-08-11
fix(oci-dependency-resolution): Support OCI resolution and empty repository for chart dependencies.

therealgambo:
This PR addresses two shortcomings in the way which dependencies are currently resolved.
- lack of `oci://` support in Chart.lock
- incorrect resolution of `repository: ""` references, where only `file://` was previously supported.

#### OCI dependency
Each dependency entry in Chart.lock gets resolved via `repo.FindChartInRepoURL`, which only understands classic HTTP(S) chart repos which serve an index.yaml file. If a dependency is referenced via `oci://`, the current mechansim appends `/index.yaml` to the OCI repo URL which produces in invalid format preventing dependencies from being downloaded during the intial "prepare" stage.

```
09:26:03 ERR charts/package.go:48 encountered error while preparing chart path=charts error="encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency postgresql: looks like \"oci://registry-1.docker.io/bitnamicharts\" is not a valid chart repository or cannot be reached: failed to perform \"FetchReference\" on source: invalid reference"
09:26:03 ERR logger/log.go:36 encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency postgresql: looks like "oci://registry-1.docker.io/bitnamicharts" is not a valid chart repository or cannot be reached: failed to perform "FetchReference" on source: invalid reference
```

Added new wrapper function `resolveDependencyURL()`, which exits early if `helmRegistry.IsOCI()` is true returning the correctly formatted URL before /index.yaml can be appended.

I discovered this limitaton in the tooling when attempting to patch the infisical-standalone chart for my side project work.
- https://github.com/Infisical/infisical/blob/main/helm-charts/infisical-standalone-postgres/Chart.yaml#L23-L30

#### Empty Repository
If a chart has a dependency with `repository: ""` set, it means the dependency is already vendored alongside the main chart under `charts/<name>`, as such we don't need to perform any additional lookups to retrieve and download that source.

Previously, the tooling only supported `file://`. This brings closer alignment to how default helm functionality works.

```
10:23:43 ERR charts/package.go:48 encountered error while preparing chart path=charts error="encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency podlogs-crd: could not find protocol handler for: "
10:23:43 ERR logger/log.go:36 encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency podlogs-crd: could not find protocol handler for:
```

See: https://github.com/grafana/alloy-operator/blob/main/charts/alloy-operator/Chart.yaml#L23

### DIFF
--- a/pkg/charts/dependencies.go
+++ b/pkg/charts/dependencies.go
@@ -20,6 +20,7 @@ import (
 	helmLoader "helm.sh/helm/v3/pkg/chart/loader"
 	helmCLI "helm.sh/helm/v3/pkg/cli"
 	helmGetter "helm.sh/helm/v3/pkg/getter"
+	helmRegistry "helm.sh/helm/v3/pkg/registry"
 	helmRepo "helm.sh/helm/v3/pkg/repo"
 	"sigs.k8s.io/yaml"
 )
@@ -113,6 +114,30 @@ func PrepareDependencies(ctx context.Context, rootFs, pkgFs billy.Filesystem, ma
 	}
 
 	return UpdateHelmMetadataWithDependencies(ctx, pkgFs, mainHelmChartPath, dependencyMap)
+}
+
+// resolveDependencyURL resolves the download URL for a single chart dependency listed in Chart.lock.
+//
+// OCI-based chart repositories do not serve an index.yaml the way classic HTTP(S) chart repositories
+// do, so helmRepo.FindChartInRepoURL cannot be used to resolve them. For OCI repositories we instead
+// build the fully-qualified "oci://host/path/chart:version" reference directly;
+// GetUpstream/puller.Registry already knows how to pull that.
+func resolveDependencyURL(repository, name, version string) (string, error) {
+	if helmRegistry.IsOCI(repository) {
+		return fmt.Sprintf("%s/%s:%s", strings.TrimSuffix(repository, "/"), name, version), nil
+	}
+
+	// Acquire mutex to serialize Helm repository queries and prevent rate limiting
+	helmRepoFetchMutex.Lock()
+	defer helmRepoFetchMutex.Unlock()
+
+	return helmRepo.FindChartInRepoURL(
+		repository,
+		name,
+		version,
+		"", "", "",
+		helmGetter.All(&helmCLI.EnvSettings{}),
+	)
 }
 
 func getMainChartUpstreamOptions(ctx context.Context, pkgFs billy.Filesystem, gcRootDir string) (*options.UpstreamOptions, error) {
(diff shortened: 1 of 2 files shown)


nicholasSUSE:
> This PR addresses two shortcomings in the way which dependencies are currently resolved.
> 
>     * lack of `oci://` support in Chart.lock
> 
>     * incorrect resolution of `repository: ""` references, where only `file://` was previously supported.
> 
> 
> #### OCI dependency
> 
> Each dependency entry in Chart.lock gets resolved via `repo.FindChartInRepoURL`, which only understands classic HTTP(S) chart repos which serve an index.yaml file. If a dependency is referenced via `oci://`, the current mechansim appends `/index.yaml` to the OCI repo URL which produces in invalid format preventing dependencies from being downloaded during the intial "prepare" stage.
> 
> ```
> 09:26:03 ERR charts/package.go:48 encountered error while preparing chart path=charts error="encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency postgresql: looks like \"oci://registry-1.docker.io/bitnamicharts\" is not a valid chart repository or cannot be reached: failed to perform \"FetchReference\" on source: invalid reference"
> 09:26:03 ERR logger/log.go:36 encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency postgresql: looks like "oci://registry-1.docker.io/bitnamicharts" is not a valid chart repository or cannot be reached: failed to perform "FetchReference" on source: invalid reference
> ```
> 
> Added new wrapper function `resolveDependencyURL()`, which exits early if `helmRegistry.IsOCI()` is true returning the correctly formatted URL before /index.yaml can be appended.
> 
> I discovered this limitaton in the tooling when attempting to patch the infisical-standalone chart for my side project work.
> 
>     * https://github.com/Infisical/infisical/blob/main/helm-charts/infisical-standalone-postgres/Chart.yaml#L23-L30
> 
> 
> #### Empty Repository
> 
> If a chart has a dependency with `repository: ""` set, it means the dependency is already vendored alongside the main chart under `charts/<name>`, as such we don't need to perform any additional lookups to retrieve and download that source.
> 
> Previously, the tooling only supported `file://`. This brings closer alignment to how default helm functionality works.
> 
> ```
> 10:23:43 ERR charts/package.go:48 encountered error while preparing chart path=charts error="encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency podlogs-crd: could not find protocol handler for: "
> 10:23:43 ERR logger/log.go:36 encountered error while trying to prepare dependencies in charts: encountered error while trying to find the repository for dependency podlogs-crd: could not find protocol handler for:
> ```
> 
> See: https://github.com/grafana/alloy-operator/blob/main/charts/alloy-operator/Chart.yaml#L23

Hey @therealgambo, I am the maintainer of `charts-build-scripts`. Sorry for the delay; I was on vacation and had a release. 
This PR looks really good. I'll review it thoroughly as soon as I finish the renovate PRs. 

Thank you for your contribution.

nicholasSUSE (approved):

